PR #14, opened 2014-10-30
use opencv's git instead of archive

D-Alex:


jakobs:
There are 3 PR open for opencv, which one (if any) is still valid?

planthaber:
This one

PR #16, opened 2014-11-10
ROCK_BUNDLE_PATH should only define super folder of bundles

maltewi:
ROCK_BUNDLE_PATH should only contain higher-level folder where bundles can be found as subfolders. It should not contain the bundle folders themselves.


maltewi:
The probem is occuring to other people as well. The suggested solution so far worked for them. Maybe merge for now and, if the solution is not quite perfect, open an issue for improvements?

doudou:
Agreed and done

PR #19, opened 2014-12-11
the opencv zip now contains a folder with version number

planthaber:


doudou:
OK ... let's merge this one. It is by far the least intrusive. In addition to the fix to make sure we install the debian package on OSes where it is supported, we should be fine.

doudou:
Then if the consensus is to delete the source package completely, then so be it.... Won't be a lot harder.

PR #21, opened 2014-12-18
use packaged "opencv" from OS

marvin2k:
previously, opencv-2.4.9 was downloaded and compiled from source. now use the precompiled, package version from the OS. this comes in:

Debian Testing (Jessie): 2.4.9
Debian Stable (Wheezy): 2.3.1  # quite old. Is Wheezy still supported? boostrap.sh fails if unmodified (`--no-check-certificate`)
Ubuntu 14.04 (Trusty): 2.4.8
Ubuntu 12.04 (Precise): 2.3.1  # 12.04 is not supported anymore?

there are more OS in the osdeps-file, where I just assume they bring similar packaged versions.

tested "image_processing/{frame_helper,stereo} in Jessie and Trusty. but obviously that shouldn't be it. which packages else?

Signed-off-by: Martin Zenzes martin.zenzes@dfki.de


jakobs:
Is this still valid?

doudou:
> Therefore, I would also vote for removing it all together because right now you cannot even overwrite external/opencv with a git in your local installation without changing the package set.

At the time, a significant number of rock installs were building opencv "by hand" because older ubuntu versions only had very old opencv versions. The problem with changing archive to git then was that it would break "autoproj update" for all of them.

@D-Alex: why would you have to modify the package set ? (as opposed to using an override ?)

D-Alex:
The last time I tried it was not possible to overwrite a tar with a git repository via overwrite.yml and I had to modify the package set

jakobs:
Seems like an issue with autoproj then.

doudou:
Just tried it. It works as expected: the importer is updated to be a git importer, but import fails because the current checkout is not a git repository

planthaber:
So could you merge this request to get the bootstrapping working until the importer issue is fixed?

doudou:
Sorry I missed something: how does this affect bootstrapping ?

planthaber:
Because the Debian definition points to the opencv source archive, inside this, the folder name changed: https://github.com/rock-core/package_set/pull/19 so bootstrapping with a package set with a opencv dependency just fails.

As the apt-get package from debian has the same version, this PR is favorable over https://github.com/rock-core/package_set/pull/19

This is not something exactly what should happen for a "supported OS", when "using it on a supported OS" can be broken vor over 2 months

doudou:
> This is not something exactly what should happen for a "supported OS", when "using it on a supported OS" can be broken vor over 2 months

Debian is not officially supported, so take a deep breath here.

There are basically two problems at hand:
 - Rock tries to build from source on jessie, while it should use the prepackaged opencv
 - the current OpenCV archive does not check out

Given that, removing the build-from-source definition is throwing the baby with the bathwater. Next time someone wants to build from source, he'll have to redo the work. I've cherry-picked and pushed the osdep commit. I would welcome changing the opencv source definition to use the git checkout (as suggested by @marvin2k) if someone is willing to do it, or I'll do it later on.

planthaber:
>I would welcome changing the opencv source definition to use the git checkout (as suggested by @marvin2k) if someone is willing to do it, or I'll do it later on.

Alex already did that:
https://github.com/rock-core/package_set/pull/14

You marked it "do not merge" because of an issue with the importer

PR #22, opened 2015-01-15
added rock-webapp ruby package

jakobs:
rock-webapp provides a way to interface with rock using a REST api. This
is e.g. for providing web based clients for rock installations.

### DIFF
--- a/00new_packages.autobuild
+++ b/00new_packages.autobuild
@@ -1,10 +1,14 @@
 # IMPORTANT: new packages must be added at the top of this file
 # They must NEVER be added to other files
 
-
 cmake_package 'tools/class_loader' do |pkg|
     pkg.depends_on 'base/console_bridge'
     pkg.depends_on 'poco'
     pkg.define "BUILD_SHARED_LIBS", "ON"
     pkg.define 'CMAKE_BUILD_TYPE', "Release"
+end
+
+cmake_package 'gui/rock_webapp'
+ruby_package 'tools/rest_api' do |pkg|
+    Autoproj.env_set 'ROCK_WEBAPP', File.join(pkg.prefix, 'share', 'rest_api')
 end
--- a/rock.osdeps
+++ b/rock.osdeps
@@ -125,3 +125,10 @@ cucumber: gem
 
 poco:
     debian,ubuntu: libpoco-dev
+
+faye-websocket: gem
+rack-cors: gem
+thin: gem
+grape: gem
+sprockets: gem
+


doudou:
I would have personally much preferred having it under tools/. To me, the web interface (i.e. the javascript stuff) is a byproduct. The important part is IMO the REST interface, to which I would like to migrate some of the underlying client/server things in the long run (syskit shell, orocos process server, ...)

jakobs:
Ok, I am fine with both, so tools is ok. Do you think rock-webapp is the right name then?

doudou:
I wondered about it as well. It would not match. Moreover, @planthaber and I decided at some point to merge the frontend and backend parts, but we might want to reconsider (i.e. split tools-rock_restapi and gui-rock_webapp)

The convention would also be to rename it with an underscore (as we decided to keep - for directory structure)

planthaber:
We merged them because the webapp ui has to be serverd from "webserver" which provides the rest API, we would run into problems with webbrowsers trying to prevent CORS (js has to be loaded from the same ip AND port, otherwise the REST API is not available fot the js code).

We could:
* allow CORS for the REST API
* Tell the "API" which additional folders (from "gui-rock_webapp") to serve

For the second solution there is a implementation to add additional resources to be served by the API, but only for a single folder. But we could extend that.

doudou:
The goal is to be able to start a single server, so I vote for solution (2). My idea would be to extend the command-line application to (1) load plugins and (2) get a list of "mount points" (that is filesystem-to-URLpath mappings)

jakobs:
ok. should it be tools-rock_restapi, tools-rest_api or tools-restapi?

doudou:
No preference, pick the one you prefer.

jakobs:
Sorry, running late on this. Btw. do should orocos.js be part of rest_api, or webapp? It might be more closely connected to webapp, but I think in a lot of cases, when you have a custom ui, this is the only reference that is needed, and the bulk of the ui part is not required. So if you are ok with it I would leave it in tools/rest_api, including the sprockets library.

doudou:
Currently, both packages are the same, right ?

What *is* orocos.js ?

jakobs:
well no, i stripped the history of the ui folder from the rest_api repo. And will remove the rb stuff from the webapp repo. Both are work in progress. Orocos.js is essentially a small refactoring of some of steffens code, to provide a small javascript library that can essentially be used like orocos.rb.

doudou:
> Orocos.js is essentially a small refactoring of some of steffens code, to provide a small javascript library that can essentially be used like orocos.rb.

Ah, alright. I don't mind you leaving it in the rest_api package

jakobs:
Ok, finally had some time to get this plugin mechanism working

planthaber:
I'm missing the source.yml entries.

jakobs:
Did you try? Should be handled by the wildcards, no?

planthaber:
didn't try, forgot about the wildcards

jakobs:
@doudou what do you think?

00new_packages.autobuild
```diff
@@ -1,4 +1,7 @@
 # IMPORTANT: new packages must be added at the top of this file
 # They must NEVER be added to other files
 
-
+cmake_package 'gui/rock_webapp'
+ruby_package 'tools/rest_api' do |pkg|
+    Autoproj.env_add_path 'ROCK_WEBAPP', File.join(pkg.prefix, 'share', 'rest_api')
```
doudou: I usually prefer to have these envvars end up with _PATH. Or is that not a PATH ? In addition, I started to think that we might want to define a ROCK_PREFIX_PATH variables from which all the other search paths are derived (the way e.g. KDE does it). That would reduce the proliferation of envvars ...

planthaber: +1 for "_PATH" 

But the ROCK_PREFIX_PATH would also make the overriding for testing harder. In case the webapp to test is in a different root folder (don't know if this case occurs often at all)

doudou: Since it is a path, you basically prepend the "test environment" at the front. This way, you get to load "test fixtures" first and fallback to global ones otherwise.

Given that it is how all the path stuff works (e.g. RUBYLIB or PYTHONPATH or KDEPREFIXES), I guess that it should not be completely broken.

planthaber: > Since it is a path, you basically prepend the "test environment" at the front. This way, you get to load "test fixtures" first and fallback to global ones otherwise.

Jep, but you have to create the very same folder structure (including subfolders), also when you could only need to use a simple checkout, which could be located just "somewhere". You cannot point to a location outside of the current istallation, I guess changing the PREFIX_PATH in rock is a bad idea in general.

I never said that it is completely broken. Both have their "plus", but I fear that if someone is not so familiar with paths, that person could try to change the prefix instead of the webapp path and "break everything" rather than just use another webapp. Normally you don't have to care about paths at all when using rock, so a lot of users might be unfamiliar with path settings.

doudou: OK ... I seem to be missing something here.

What is this variable used for exactly ?

planthaber: The variable is used to link "tools/rest_api" to the html part in "gui/webapp", as the html part hat to be provided by the rest_api on the same port as the api in order to be able to use the api with JS without the need to enable CORS (I hope you remember).

The variable is meant to hold several paths to html/js pages meant to access the rest api. It should be extenable for specific pages for single robots (like robot specific control UIs). So the variable holds a list of paths.

I think the point of discussion is about how to use the ROCK_PREFIX_PATH proposed by you.

IMHO it should be evaluated on varaible (environent) level and not on source level:

export ROCK_WEBAPP_PATH=${ROCK_PREFIX_PATH}/gui/webapp:${ROCK_PREFIX_PATH}/gui/ASIMO/webapp

This way custom (global) paths can be added as well as the prefix path can be used and changed.

jakobs: Sorry, but this is not what it is anymore. Currently ROCK_WEBAPP (still without the path, but have no problem changing it) points to install/share/rest_api. Rock-webapp will look for any *.rb files in install/share/rest_api/plugin. Inside these plugins, you can modify the rack mappings, and add search paths for sprockets (which is used to serve and preprocess js and css). The way I currently did it that the mapping for e.g. for the webapp would point to the public directory of the webapp source package (or whichever other package).

jakobs: Also, ROCK_WEBAPP is a single path, and does not support having multiple paths at the moment. This could be added later, but may not even be needed.

doudou: OK then. If it's not a path, I'm all OK with the current envvar name. Only, don't use env_add_path but env_set. Migration to a hypothetical ROCK_PREFIX_PATH would obviously require too much time to be relevant here anyways. Was just a point for future discussions.

doudou:
Oh ... and it does not merge automatically anymore ... :P

jakobs:
squashed and rebased.

PR #25, opened 2015-02-20
class_loader: mark as stable

goldhoorn:
Otherwise the debug version of libpoco is slected with might be not installed


doudou:
I would rather explicitely set the build type as Release, as it is what is actually needed (see #26)

goldhoorn:
Ok

PR #26, opened 2015-02-20
build class_loader in release mode

doudou:
This workarounds an Ubuntu bug on Ubuntu 14.10, in which the symlinks
to the debug version of pocoFoundation are present but not the actual
libs.

The other option would be to install the -dbg package.


goldhoorn:
Ack for this

PR #31, opened 2015-03-16
add the thor osdep

doudou:
Thor is a [CLI-building tool](whatisthor.com). I've started to use it for some tooling instead of OptionParser because it feels a lot cleaner. You really build an API and tell thor how to expose it to the command line, instead of having a big toplevel script.


goldhoorn:
ack

doudou:
Actually, this got added by #30

PR #35, opened 2015-03-26
Small osdeps fixes

doudou:


rock.osdeps
```diff
@@ -136,3 +138,10 @@ thin: gem
 grape: gem
 sprockets: gem
 
+qtruby:
+    default:
+        gem: qtbindings
+        osdep: [qt4, qt4-opengl, qt-designer, qt4-webkit]
```
goldhoorn: Do we really need the designer for people who dosn't want to design her own guis?

doudou: No, but so far these are designed with the intent of getting a developer environment ready for people. In addition, this PR is really about doing the same-than-before. If you want to change the status quo, best do it in another PR.

D-Alex: the designer is installing some dev-libraries needed for the vizkit ui loading mechanism

goldhoorn:
then ack for merge

PR #37, opened 2015-04-02
tune testing support

doudou:
This controls ROCK_TEST_ENABLED (from the Rock cmake macros) based on whether
the underlying package has tests enabled or not, and makes sure that we build
orocos.rb's test oroGen components during the build phase if tests are enabled
for it (and that we do so with the appropriate -j)

### DIFF
--- a/04stable.autobuild
+++ b/04stable.autobuild
@@ -125,6 +125,18 @@ in_flavor 'master', 'stable' do
             Autobuild.update_environment srcdir
             Autoproj.env_set 'ORBInitRef',"NameService=corbaname::127.0.0.1"
         end
+
+        def pkg.invoke_rake(target_name = rake_setup_task)
+            if test_utility.enabled?
+                Autobuild.invoke_make_parallel(self) do |*make_options|
+                    make_options = make_options.
+                        map { |opt| opt.gsub(',', ';') }
+                    test_target = "setup:orogen_all[1,#{Autobuild::Orogen.transports.join(" ")},'#{make_options.join("' '")}']"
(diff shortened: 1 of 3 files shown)


doudou:
Ping @rock-core/core

doudou:
Ok ... I can see that unit testing is very high in the priority lists.

04stable.autobuild
```diff
@@ -120,10 +120,16 @@ in_flavor 'master', 'stable' do
     end
 
     ruby_package  'tools/orocos.rb' do |pkg|
-        def pkg.update_environment
+        def pkg.invoke_rake(target_name = rake_setup_task)
             super
-            Autobuild.update_environment srcdir
-            Autoproj.env_set 'ORBInitRef',"NameService=corbaname::127.0.0.1"
```
goldhoorn: How is this removal related to test support?

doudou: It's not ... updated

goldhoorn:
As far i understood this enables the build of all test-based extensions by default.
Furthermore this is done also for stable and not only for master.

I would prefer to have a kind of ENV or Autoproj setting that controls if the test suites are build by default instead of activating them automatically.

I have also questions regarding testing in rock in general:

this PR will enable the compilation of all test parts by default
autoproj test will run the tests?
why call autoproj osdeps after autoproj test?

goldhoorn:
Ah okay get it. ```autoproj test --enable``` enables the test_utility for each package. 
The tests got therefore not build by default. Only if this is set by autoproj configuration

PR #40, opened 2015-04-15
rock.core package set affected by renaming image_processing 

jhidalgocarrio:
It aims to rename image_processing to perception (core packages affected). Discussed in the rock-dev mailing list email.

It should be done in conjunction with https://github.com/rock-core/rock-package_set/pull/53


doudou:
Same comment than for rock-package_set#53: add metapackage lines to keep backward compatibility

goldhoorn:
all one, working under old and new location

PR #41, opened 2015-04-21
Update Arch Linux osdeps

phyrog:
The osdeps for Arch Linux are partially outdated. This PR gets it up and running on Arch again.

### DIFF
--- a/06osdeps_from_source.autobuild
+++ b/06osdeps_from_source.autobuild
@@ -14,5 +14,9 @@ in_flavor 'master', 'stable' do
     end
     Autoproj.add_osdeps_overrides 'opencv', :package => 'external/opencv'
     remove_from_default 'external/opencv'
+
+    cmake_package 'tools/poco'
+    Autoproj.add_osdeps_overrides 'poco', :package => 'tools/poco'
+    remove_from_default 'tools/poco'
 end
 
--- a/manifests/tools/poco.xml
+++ b/manifests/tools/poco.xml
@@ -1,0 +1,9 @@
+<package>
+  <description brief="C++ class libraries for network-centric, portable applications">
+    Modern, powerful open source C++ class libraries for building network- and internet-based applications that run on desktop, server, mobile and embedded systems.
+  </description>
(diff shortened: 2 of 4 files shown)


06osdeps_from_source.autobuild
```diff
@@ -14,5 +14,9 @@ in_flavor 'master', 'stable' do
     end
     Autoproj.add_osdeps_overrides 'opencv', :package => 'external/opencv'
     remove_from_default 'external/opencv'
+
+    cmake_package 'external/poco'
+    Autoproj.add_osdeps_overrides 'poco', :package => 'external/poco'
```
goldhoorn: we decided a while ago (which you cannot know) that the external/ folder should not be used anymore.
external poco is used by class_loader. I recomment to place the poco inside tools/

source-stable.yml
```diff
@@ -13,6 +13,11 @@ version_control:
       type: archive
       url: http://downloads.sourceforge.net/project/opencvlibrary/opencv-unix/2.3.1/OpenCV-2.3.1a.tar.bz2
       update_cached_file: false
+    - external/poco:
+      type: archive
+      url: https://github.com/pocoproject/poco/archive/poco-1.6.0-release.zip
+      update_cached_file: false
+      archive_dir: poco-poco-1.6.0-release
```
goldhoorn: You don't need to define this twice if this is the same version. the stable is only if on "stable" a different version then on master is used.

06osdeps_from_source.autobuild
phyrog: Fixed.

source-stable.yml
phyrog: Fixed too.

goldhoorn:
You missed to rename the manifest.xml file ;)

phyrog:
Oops, fixed too.

doudou:
Hey. The qtruby osdeps changed since the merge of #35 ... Could you update the PR ?

phyrog:
@doudou Sure, done.

PR #44, opened 2015-05-07
fix URLs

mintar:


mintar:
Both "http://rock-robotics.org/release/documentation/..." and "http://rock-robotics.org/documentation/..." work; I wasn't sure which one to use, so I took the first one.

goldhoorn:
stable will come back, currently we fighting with some inconsistencies, there is currently no need to updated the doc, but thanks for assisting

mintar:
Ok, good to know! I'll close this PR then.

PR #48, opened 2015-05-26
remove mqueue for Mac OS X from default configuration

malter:


doudou:
I just merged without giving it much thought ... However, these are POSIX message queues. What's wrong with them on MacOSX ?

malter:
I didn't dig into it, but it seems to be the default behavior of rtt to build without mqueue support:
config/check_depend.cmake:  message("Forcing ENABLE_MQ to OFF for macsox")

PR #49, opened 2015-05-27
Changed vizkit plugin name to meet repro name

goldhoorn:


goldhoorn:
Must applied together with:
https://github.com/rock-core/base-scripts/pull/19

PR #52, opened 2015-05-28
added mac ports information for system dependencies

malter:


goldhoorn:
@D-Alex: can you please check is avahi is really not availible for brew?

goldhoorn:
Otherwise if you not sure i prefer no entry because new users might get a error instead simply ignoring it that avahi is not installed (which is the better behaviour (for me))

goldhoorn:
Valid not no impact for other package managers...

PR #53, opened 2015-05-29
fixed wrong eigen packages for macos-port

malter:


goldhoorn:
Next time please check first before doing a PR ;)

PR #54, opened 2015-05-29
Reworked exclusion of OCL if USE_OCL is false

goldhoorn:


goldhoorn:
OCL is unable to compile without USE_OCL is true. 
If it tries to get compile without OCL support it will fail because RTT got compile without OCL compatibility. Therefore a hard exclusion is valid. and if it got selected a hard error is thrown.

PR #55, opened 2015-06-04
merge date 10.06.2015: Removed ruby1.9 support

goldhoorn:
Remove compatibility to 1.9

### DIFF
--- a/init.rb
+++ b/init.rb
@@ -12,12 +12,13 @@ ignore(/\.sw?$/)
 # Ignore the numerous backup files
 ignore(/~$/)
 
-# Verify that Ruby is NOT 1.8. 1.8 is unsupported
+
+# Ruby 1.8 is completly outdated, if you modify this, take respect to the addition checks below against 1.9 
 if defined?(RUBY_VERSION) && (RUBY_VERSION =~ /^1\.8\./)
     Autoproj.error "Ruby 1.8 is not supported by Rock anymore"
     Autoproj.error ""
     Autoproj.error "Use Rock's bootstrap.sh script to install Rock"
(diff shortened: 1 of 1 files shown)


goldhoorn:
This should not merged before 10.05.2015

goldhoorn:
Btw. 
why do we have this bootstrap.sh scripts at all i dont like them. 
They are copied around, and from a maintainer sight we have no way to adapt them.
The branch settings get more complex. Why we not simply tell the customers howto bootstrap with using the autoproj_bootstrap script and removing this file at all?

goldhoorn:
wah wrong PR (again) sorry

init.rb
```diff
@@ -12,15 +12,6 @@
 # Ignore the numerous backup files
 ignore(/~$/)
 
-# Verify that Ruby is NOT 1.8. 1.8 is unsupported
-if defined?(RUBY_VERSION) && (RUBY_VERSION =~ /^1\.8\./)
```
doudou: Should keep this for stable, no ?

goldhoorn:
*updated*

I will apply this patch on 10.06.2015 as discussed

init.rb
```diff
@@ -12,12 +12,13 @@
 # Ignore the numerous backup files
 ignore(/~$/)
 
-# Verify that Ruby is NOT 1.8. 1.8 is unsupported
+
+# Ruby 1.8 is completly outdated, if you modify this, take respect to the addition checks below against 1.9 
 if defined?(RUBY_VERSION) && (RUBY_VERSION =~ /^1\.8\./)
```
2maz: Looks like this section can be removed, since the next is checking on Ruby < 2.0

goldhoorn: Is is still needed because stable can be still 1.9

PR #60, opened 2015-07-06
removed gitorious package settings

planthaber:
Apart from base/templates/doc, all of them are should now be fetched by the regexp
source definition.

I am unsure about base/templates/doc, which pointed to doc_preview on
gitorious, github has both templates-doc and templates-doc_preview.


planthaber:
Similar patch megred from other brenach

PR #63, opened 2015-08-05
Corrected the base/templates/doc repository from gitorious to github.

rafaelsaback:
Corrected the base/templates/doc repository from gitorious to github.


doudou:
You should not use the rock-rc branch. Ever.

Actually, we should delete all rock-rc branches.

PR #66, opened 2015-08-24
Define some gems

doudou:


doudou:
This supersedes #65

2maz:
merged

PR #68, opened 2015-09-23
Use in Ubuntu the repository version of the ruby package flexmock

Rauldg:
This at least solved the issue for my installation.

### DIFF
--- a/rock.osdeps-ruby19
+++ b/rock.osdeps-ruby19
@@ -3,3 +3,5 @@ rake-compiler:
         gem: rake-compiler<0.9.0
     default:
         gem: rake-compiler
+flexmock:
+    gem: "flexmock<2.0"


doudou:
> "The gem version requires ruby 2.0 or later"

So does Rock ... so ?

jmachowinski:
The problem is, that you can not bootstrap older releases any more using ruby 1.9

Rauldg:
Shouldn't the stable branch be ruby1.9 compatible? So far I was assuming that. In fact I thought ruby2 was not (at least officially) supported for the stable branch.

Why is there no stable branch for this repository... Doesn' it make sense?

goldhoorn:
The package sets are not flavored, if you need to use a speficic packageset flavor behaviour use the foo.stable definitions

doudou:
@Rauldg: oooops. You are absolutely right.

The best fix is to add

~~~
flexmock:
   gem: "flexmock<2.0"
~~~

in rock.osdeps-ruby19. It's nice because doing a point release is then easy. The long term fixes are to add flexmock only as test_depend (I've done it for utilrb), and to replace rubygems by bundler in autoproj.

Rauldg:
Thanks! 
@doudou I applied the change and tested. 
Should I squeeze the two commits before merge?

doudou:
> Should I squeeze the two commits before merge?

Yes, please.

Rauldg:
^.^ Done!

PR #70, opened 2016-01-29
set pastel version

planthaber:
Not defining it specificly results in following error:

/usr/lib/ruby/2.1.0/rubygems/specification.rb:2064:in `raise_if_conflicts': Unable to activate tty-0.3.2, because pastel-0.6.0 conflicts with pastel (~> 0.5.1) (Gem::LoadError)


planthaber:
added activesupport version, gem currently installs v5, which requires ruby > 2.2

PR #71, opened 2016-02-08
install os package of activesupport which should be compatible with t…

planthaber:
…he current os ruby version

the current gem requires ruby 2.2

### DIFF
--- a/rock.osdeps
+++ b/rock.osdeps
@@ -34,7 +34,10 @@ binding_of_caller:
     gem: binding_of_caller
 
 activesupport:
-    gem: activesupport <= 4.2.5.1
+  ubuntu: ruby-activesupport
+  debian: ruby-activesupport
+  default:
+    gem: activesupport
 
 eigen3:
     debian, ubuntu: libeigen3-dev
@@ -162,6 +165,7 @@ faye-websocket: gem
 rack-cors: gem
 thin: gem
 grape: gem
+
 sprockets: gem
 
 thor:


planthaber:
any comments?

rock.osdeps
```diff
@@ -34,7 +34,10 @@ binding_of_caller:
     gem: binding_of_caller
 
 activesupport:
-    gem: activesupport <= 4.2.5.1
+  ubuntu: ruby-activesupport
+  debian: ruby-activesupport
+  default:
+    gem: activesupport
```
doudou: Why have you removed the version constraint here ?

planthaber: the current activesupport 5.0 is only compatible to ruby >= 2.2 

Installing the osdep is just the cleaner fix for the issue rather than adding a version contraint.
https://github.com/rock-core/package_set/pull/70


The osdep should always fit the ruby version installed on the system.

For the gem version you can't tell whether someone uses an older ruby, or has the current version installed and wants to install the latest version of activesupport. Not having the version dependency is more future proof than fixing the version. 

Also because one tends to forget why the version constraint was introduced.

doudou: Then the right fix is to have a per-ruby-version definition of activesupport (look at the rock.osdeps-ruby* files). Many people don't use the prepackaged version of Ruby.

planthaber: That's right, but the grape gem also has activesupport in its dependency list (unversioned), on new bootstraps on the buildserver it happens that the grape gem is evaluated fist and tries to install the current version of activesupport via the gem dependency mechanism, adding it as osdep it makes sure that activesupport is already installed, when the grape gems dependencies are evaluated.

Or is there a way to tell autoproj to install specific gems first?

doudou: There are no ways to deal with that, really. Autoproj 2.0 fixes it by using bundler.

planthaber: well, except installing them as debian packages

doudou:
OK ... there is indeed no good solution (until autoproj 2.0 gets around)

PR #76, opened 2016-03-21
use OS packaged version if the hoe gem

planthaber:
Hoe gem fails recently because of version conflicts with rake:

on master:
in directory /data/jenkins/workspace/workspace/rock-basic/Flavor/master/node/Ubuntu_15.04/dev/base/scripts
    cannot load the Hoe gem. Distribution is disabled

on stable:
'utilrb' cannot be build -- loading gem failed: Unable to activate hoe-3.14.2, because rake-11.1.1 conflicts with rake (< 11.0, >= 0.8)

using the packages version solves the issue

### DIFF
--- a/rock.osdeps
+++ b/rock.osdeps
@@ -60,7 +60,12 @@ webgen:
         - cmdparse<3.0
 coderay: gem
 kramdown: gem
-hoe: gem
+hoe: 
+    ubuntu:
+      '14.04,14.10,15.04': ruby-hoe
+      default: gem
+    default: gem
+
 yard: gem
 hoe-yard: gem
 rice:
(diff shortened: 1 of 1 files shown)


jmachowinski:
is this commit still needed ?

planthaber:
I guess so, the problem is that a lot of current versions gems are incompatible to ruby 2.0, thisd also applies to e.g. rake, as long autoproj installs them this is fine as you can set the version to be installed, in case it is installed via rubygems dependency gem install takes the current version, no matter if compatible or not.

The only way around this would be to install the gem with the lower version BEFORE the dependency mechanism of gem install tries to install it (gem already available). In autoproj 1.9 there is no way to do this than using the debian packages, which should always be compatible . 

long story short: the issue was with rake , so the rake-compiler debian package should be installed to fix this.

PR #77, opened 2016-03-29
remove usage of APIs deprecated in autoproj 2.0 in a 1.x-compatible way

doudou:
This uses API that was already present in 1.x, but for which no
deprecation warnings had been issued. 2.x will be pretty verbose
about these, so change the autobuild code to either use the new
APIs, or be compatible with both 1.x and 2.0

### DIFF
--- a/04stable.autobuild
+++ b/04stable.autobuild
@@ -1,11 +1,16 @@
+manifest = if Autoproj.respond_to?(:workspace)
+               Autoproj.workspace.manifest
+           else Autoproj.manifest
+           end
+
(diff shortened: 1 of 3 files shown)


doudou:
Anyone in @rock-core/core ? Also https://github.com/orocos-toolchain/autoproj/pull/18 and https://github.com/rock-core/rock-package_set/pull/111

2maz:
This and the related commits for the other package sets are ok -- tested on 14.04, autoproj v1.13.4, autobuild 1.9.3.

PR #81, opened 2016-07-04
extract the Eigen ruby bindings in a separate package

doudou:
This will allow releasing them as a gem, thus improving the integration
with both ruby-doc and travis-ci for this package and for other pure-ruby
packages that depend on it (but not on base/types)

Having it installed at the same time than base/types should be
harmless (it will pick one or the other depending on the setup of
RUBYLIB). The bindings should be removed from base/types once this
is merged.

The library has had its test suite extended, which uncovered a few
bugs. Documentation has been improved dramatically by documenting
the C++ extension.

### DIFF
--- a/00new_packages.autobuild
+++ b/00new_packages.autobuild
@@ -13,3 +13,5 @@ cmake_package 'gui/rock_webapp'
 ruby_package 'tools/rest_api' do |pkg|
     Autoproj.env_set 'ROCK_WEBAPP', File.join(pkg.prefix, 'share', 'rest_api')
 end
+
+ruby_package 'base/ruby_eigen'


doudou:
From what it's worth, an alternative would be to directly use [the gem](https://rubygems.org/gems/eigen)

jmachowinski:
If we are going down the road to the gem, what about using 
https://github.com/ruby-eigen/ruby-eigen
instead ? It looks a lot more feature complete.

doudou:
> what about using  https://github.com/ruby-eigen/ruby-eigen instead ? It looks a lot more feature complete.

Generally speaking, it sounds like a good idea. However:
 - it's using SWIG. Meaning that I would NOT use this gem without a LOT of testing beforehand (last time I used swig-generated bindings, memory management was VERY flaky)
 - it has an incompatible API, meaning that I would have to write an adapter.

If I have to anyway evaluate a new gem and write adapter APIs, I would definitely prefer migrating to sciruby instead, which will most probably be supported by jruby and definitely be more supported by the community at large.

doudou:
Ping

jmachowinski:
Your call, I won't be working on it...

jakobs:
I don't get the comment with sciruby. AFAK they don't provide bindings for eigen, but have their own linear algebra package. 

My oppinion on separation: It depends on what you think the future of the eigen bindings should be. In general I think it would be ok to have them as a gem. But I feel it should only be done if there are plans for these bindings.

doudou:
The comment on sciruby: I don't see why we *have to* use Eigen on the Ruby side. sciruby aims at providing the same services (and some) for Ruby, the way the scientific python guys use their algebra. In the end, it ensures that it is actually maintained by the community, instead of having to rely on our own maintenance of very limited Ruby bindings for Eigen.

> But I feel it should only be done if there are plans for these bindings.

That's where I disagree. It basically extracts the gem in a place where it would be easier to "separate" from rock.core in the long run, should we go one way or the other.

doudou:
@jakobs: ping ?

Issue #83, opened 2016-07-11
Change rock.osdep of vtk-qt4 for ubuntu 16.04


chhtz:
Due to this change in rock-core-rock: https://github.com/rock-core/rock-package_set/commit/5f02bb99ef9fa78f3aa82d267ac82d3a577cb9f0
the os-dependency `vtk-qt4` should be changed to `libvtk6-qt4-dev` for ubuntu 16.04+.

jmachowinski:
Actually it should already be libvtk6-dev

chhtz:
Changing it for ubuntu 14.04 is another thing -- but at least libvtk and libvtk-qt4 must be the same version (ubuntu refuses to install incompatible versions of these).

jmachowinski:
Ahh I got confuses by vtk and vtk-qt4, which are two different things...

jmachowinski:
Okay, this issue is a lot worse than it looks. There is no qt4 version for vtk6. So we must either revert back to vtk5-dev or switch to qt5.

chhtz:
There is a libvtk6-qt4-dev in ubuntu 16.04. So just changing libvtk-qt for 16.04 would do the trick (i.e., leaving the libvtk-dev in rock-core/rock-package_set as it is). I don't know why these dependencies are in different package-sets to begin with, though.

jmachowinski:
There is a libvtk6-qt-dev, but that one links against qt5.

chhtz:
I think #101 essentially fixed this (at least for me).

chhtz:
Hm, there still seems to be a problem, when using the package `slam/pcl` instead of the OS version, because `vtk-qt4` is in https://github.com/rock-core/rock-package_set/blob/master/manifests/slam/pcl.xml

chhtz:
Problem can be reproduced by adding this line into your project's `override.rb`:
```
Autoproj.add_osdeps_overrides 'slam/pcl', :force => true
```
(This is necessary, if one requires another version of PCL than the OS provides)

jmachowinski:
Hm, vtk is optional, so making the dependency optional should do the trick

jmachowinski:
Made it optional, build works again

PR #84, opened 2016-07-11
added missing dep for Ubuntu 16.04

planthaber:


jmachowinski:
Please look at  rock-core/package_set#83
and 
rock-core/rock-package_set#114

planthaber:
This should be compatible, I had no linker errors with libpcl...

PR #85, opened 2016-07-12
added configuration option for xml importer

hemker:
I'm currently trying to build rock on Ubuntu 16.04 in master flavor. The default xml importer gccxml is not working with base/orogen/std. Due to a mailing list issue I added an autoproj configuration option.

### DIFF
--- a/init.rb
+++ b/init.rb
@@ -48,6 +48,18 @@ Rock.flavors.select_current_flavor_by_name(
 
 current_flavor = Rock.flavors.current_flavor
 
+#The following option is used to determine which xml importer should be used.
+#Currently castxml is working
+
+configuration_option('TYPELIB_CXX_LOADER', 'string',
+    :default => 'castxml',
+    :possible_values => ['gccxml', 'castxml'],
+    :doc => [
+        "Which xml importer should be used?",
+        "Currently only castxml is working when using master flavor"])
+
+Autoproj.env_set 'TYPELIB_CXX_LOADER', Autoproj.config.get('TYPELIB_CXX_LOADER')
+
 #This check is needed because the overrides file will override the FLAVOR selection.
 #Furthermore a selection != stable can cause a inconsistent layout (cause by in_flavor system in the package_sets)
 if File.exists?(File.join(Autoproj.root_dir, "autoproj", "overrides.d", "25-release.yml")) && current_flavor.branch != "stable" 


init.rb
```diff
@@ -48,6 +48,18 @@
 
 current_flavor = Rock.flavors.current_flavor
 
+#The following option is used to determine which xml importer should be used.
+#Currently castxml is working
+
+configuration_option('TYPELIB_CXX_LOADER', 'string',
+    :default => 'castxml',
```
doudou: Keyword argument syntax is `key: value`. `:key => value` only works as a backward compatibility feature.

init.rb
```diff
@@ -48,6 +48,18 @@
...
+    :possible_values => ['gccxml', 'castxml'],
+    :doc => [
+        "Which xml importer should be used?",
+        "Currently only castxml is working when using master flavor"])
```
doudou: This is not true. It is only required on gcc5 machines. One way could be to try and detect whether gcc is gcc5 and select gccxml/castxml on this basis.

jmachowinski: Actually it is required, as soon as you use C++11 features in transport types, or if you use gcc5

doudou: Well ... then I'm missing the point of letting it be a configuration option. If we require C++11 in the next release and in master, then starting with master and the next release, C++11 should be enabled unconditionally. Let's not burden the whole thing with optional behaviours ....

jmachowinski: Then we need to drop 14.04. CastXML is missing a patch in the osdep version on 14.04

doudou: Well. The next release should definitely not support 14.04 IMO.

Can castxml be built on 14.04 ? It does not exist as ubuntu package there, and the llvm version is too old.

jmachowinski: Interesting, I thought the 14.04 support requirement came from you. Who brought it up then ? Do we have a rule about this or something ?

doudou: We are supposed to keep support for the old LTS until LTS+1 is out (i.e. 16.10)

Given the timeline, I would either (1) decide to drop that rule or (2) release master as-is RIGHT NOW and re-release with C++11 support at the end of the year.

2maz:
What is wrong with using Autoproj's operating system detection for handling the default case, and conditionally activate gcc (14.04) or castxml (16.04)?

doudou:
> What is wrong with using Autoproj's operating system detection for handling the default case, and conditionally activate gcc (14.04) or castxml (16.04)?

Something almost equivalent, but closer to the actual problem (and therefore probably more explicit) would be to check for castxml's clang availability (it depends on a libclang-castxml osdep). It would allow to extend to other OSes easily (e.g. macosx) by making sure that libclang-castxml is available.

Unfortunately, it requires an autoproj 1/autoproj 2 specific tests, which would be on autoproj 2 (untested !)

Note: typelib setup is handled in orocos-toolchain, that's probably where this should be done.

```
os_resolver = 
    if Autoproj.manifest.respond_to?(:os_package_resolver) # v2
        Autoproj.manifest.os_package_resolver
    else
        Autoproj.manifest.osdeps
    end

if os_resolver.has?('libclang-castxml')
    # enable castxml
else
    # enable gccxml
end
```

An IMO better option would be to check whether gccxml is available from the osdeps

PR #86, opened 2016-07-13
Added configure option, if everything should be build with c++11

jmachowinski:
Note, as env_add creates a : between all elements, I needed a new helper method.
This should go into autoproj itself in the future.

### DIFF
--- a/init.rb
+++ b/init.rb
@@ -12,7 +12,6 @@ ignore(/\.sw?$/)
 # Ignore the numerous backup files
 ignore(/~$/)
 
-
 # Ruby 1.8 is completly outdated, if you modify this, take respect to the addition checks below against 1.9 
 if defined?(RUBY_VERSION) && (RUBY_VERSION =~ /^1\.8\./)
     Autoproj.error "Ruby 1.8 is not supported by Rock anymore"
@@ -35,16 +34,16 @@ Rock.flavors.define 'stable'
 Rock.flavors.alias 'stable', 'next'
 Rock.flavors.define 'master', :implicit => true
 
-def add_compiler_flag(varName, value)
-    curVal = Autobuild.env_value(varName)
-    if(curVal)
-        curVal = curVal.join(' ')
-        if(!curVal.include?(value))
-            value.concat(' ')
-            value.concat(curVal)
+def add_compiler_flag(var_name, value)
+    cur_val = Autobuild.env_value var_name
+    if cur_val
+        cur_val = cur_val.join ' '
+        if !cur_val.include? value
+            value.concat ' '
+            value.concat cur_val
         end
     end
-    Autobuild.env_set(varName, value) 
+    Autobuild.env_set var_name, value
 end
 
 configuration_option('ROCK_SELECTED_FLAVOR', 'string',
@@ -64,8 +63,8 @@ Autoproj.configuration_option 'cxx11', 'boolean',
 :default => 'yes',
 :doc => ["Compile with Cxx11 ? [yes/no]"]
 
-if (Autoproj.user_config('cxx11')) then
-    add_compiler_flag('CXXFLAGS', "-std=c++11")
+if Autoproj.user_config 'cxx11'
+    add_compiler_flag 'CXXFLAGS', "-std=c++11"
 end
 
 #This check is needed because the overrides file will override the FLAVOR selection.
--- a/init.rb
+++ b/init.rb
@@ -35,6 +35,18 @@ Rock.flavors.define 'stable'
 Rock.flavors.alias 'stable', 'next'
 Rock.flavors.define 'master', :implicit => true
 
+def add_compiler_flag(varName, value)
+    curVal = Autobuild.env_value(varName)
+    if(curVal)
+        curVal = curVal.join(' ')
+        if(!curVal.include?(value))
+            value.concat(' ')
+            value.concat(curVal)
+        end
+    end
+    Autobuild.env_set(varName, value) 
+end
+
 configuration_option('ROCK_SELECTED_FLAVOR', 'string',
     :default => 'stable',
     :possible_values => ['stable', 'master'],
@@ -47,6 +59,14 @@ Rock.flavors.select_current_flavor_by_name(
     ENV['ROCK_FORCE_FLAVOR'] || Autoproj.config.get('ROCK_SELECTED_FLAVOR'))
 
 current_flavor = Rock.flavors.current_flavor
+
+Autoproj.configuration_option 'cxx11', 'boolean',
+:default => 'yes',
+:doc => ["Compile with Cxx11 ? [yes/no]"]
+
+if (Autoproj.user_config('cxx11')) then
+    add_compiler_flag('CXXFLAGS', "-std=c++11")
+end
 
 #This check is needed because the overrides file will override the FLAVOR selection.
 #Furthermore a selection != stable can cause a inconsistent layout (cause by in_flavor system in the package_sets)
--- a/overrides.rb
+++ b/overrides.rb
@@ -114,4 +114,11 @@ Autoproj.post_import do |pkg|
    end
 end
 
+if Autoproj.user_config 'cxx11'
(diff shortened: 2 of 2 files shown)


init.rb
```diff
@@ -35,6 +35,18 @@
 Rock.flavors.alias 'stable', 'next'
 Rock.flavors.define 'master', :implicit => true
 
+def add_compiler_flag(varName, value)
+    curVal = Autobuild.env_value(varName)
```
doudou: please stick to a ruby style. 

- no () around ifs
- variables are snake_case
- keyword parameters are `key: value`, not `:key => value` (that's "new")

init.rb
```diff
@@ -48,6 +60,14 @@
 
 current_flavor = Rock.flavors.current_flavor
 
+Autoproj.configuration_option 'cxx11', 'boolean',
+:default => 'yes',
+:doc => ["Compile with Cxx11 ? [yes/no]"]
+
+if (Autoproj.user_config('cxx11')) then
+    add_compiler_flag('CXXFLAGS', "-std=c++11")
```
doudou: This should be in the rock-core package set IMO.

Given that the Rock macros already have a C++11 (CMake) option, it should also be set for them. The support for it is already in rock.core.

jmachowinski: > This should be in the rock-core package set IMO.

? This is the rock-core package set

init.rb
jmachowinski: > keyword parameters are key: value, not :key => value (that's "new")

Does not work on Ruby 1.9 e.g. Ubuntu 14.04

init.rb
doudou: Yes. Indeed ... My mistake.

doudou:
Just squash all the commits together. Or at least the "changed coding convention", which is really just noise.

init.rb
```diff
@@ -35,6 +34,18 @@
 Rock.flavors.alias 'stable', 'next'
 Rock.flavors.define 'master', :implicit => true
 
+def add_compiler_flag(var_name, value)
+    cur_val = Autobuild.env_value var_name
+    if(cur_val)
+        cur_val = cur_val.join ' '
+        if(!cur_val.include? value)
```
doudou: `if !cur_val.include?(value)`

init.rb
```diff
@@ -35,6 +34,18 @@
 Rock.flavors.alias 'stable', 'next'
 Rock.flavors.define 'master', :implicit => true
 
+def add_compiler_flag(var_name, value)
+    cur_val = Autobuild.env_value var_name
+    if(cur_val)
```
doudou: if cur_val

init.rb
```diff
@@ -48,6 +59,14 @@
 
 current_flavor = Rock.flavors.current_flavor
 
+Autoproj.configuration_option 'cxx11', 'boolean',
+:default => 'yes',
```
doudou: `default: 'yes'`

doudou:
Actually ... Since the consensus *is* to require C++11 in master and starting with the next release, why make it an option *at all* ? There are already too many optional things.

jmachowinski:
This affects Stable

init.rb
jmachowinski: > default: 'yes'

Still not working on Ruby 1.9...

hemker:
qtbindings is not building with cpp11 activated!

doudou:
> This affects Stable

This affects /current/ stable. So ... why not:
 - turn on C++11 on master unconditionally
 - not ask on stable. Current stable users already have (and probably already use) the envvar method to use C++11.
 - turn on C++11 unconditionally in the next release's setup
 - once the next release is out, turn on C++11 on both master and stable.

doudou:
> qtbindings is not building with cpp11 activated!

I believe that turning C++11 with CXXFLAGS is counter-productive. Let's turn it on for the Rock packages and other, non-rock developers, turn it on for their packages if they choose to do so (e.g. the way gazebo already does).

jmachowinski:
Not building the whole release with c++11 sounds like a bad idea to me. We recently ran into some ABI issues with that (e.g. OMPL/Vizkit). On 16.04 this should 'only' result in linker errors.  
The QTBindings thing is a real show stopper, and I'm a bit puzzled how to solve it.

doudou:
> On 16.04 this should 'only' result in linker errors

Since 16.04 prebuilt packages are not all built with C++11 anyways, we're not "adding" a problem.

One issue is that Qt4's moc is not compatible with C++11 anyways, so I'm really not sure that we can enable C++11 for all packages, let alone qtbindings.

planthaber:
Whouldn't this set cxx11 for each package, leaving out the gems?:

```ruby
Autobuild::Package.each do |name, pkg|
        pkg.add_compiler_flag 'CXXFLAGS', "-std=c++11"
end
```
The add_compiler_flag method has to mode to the Package class then, just using

pkg.define 'CXXFLAGS', "-std=c++11"

would override the original flags

doudou:
As far as I remember, CXXFLAGS is not a cmake define (i.e. it has no effect), but I might be wrong. Out of the top of my head, CMAKE_CXX_FLAGS is the right option.

> The add_compiler_flag method has to move to the Package class then, just using

I'm against adding yet-another-specific-method (which would have to be forward and back ported to autoproj v1 and v2, and wait for everyone to use the "right" autoproj version(s))

The following would be better, and can be merged with other "mandatory defines" already existing in rock.core/overrides.rb

```
Autoproj.post_import do |pkg|
    if pkg.kind_of?(Autobuild::CMake)
       pkg.define "CXXFLAGS", "#{pkg.defines['CXXFLAGS']} -std=c++11"
    end
end
```

planthaber:
Not all packages are build using cmake, but most are using "make". CXXFLAGS is the environment variable for seeting it on make level.

But you are right, when using any CMAKE_CXX_FLAGS, it completely overrides the CXXFLAGS previously set in the environment (saw that issue when we set the use of gold linker by CXXFLAGS).

Nevertheless, this solutions seems to work so far:

```ruby
Autobuild::Package.each do |name, pkg|
    if pkg.kind_of?(Autobuild::CMake)
        cxx_flags = "#{pkg.defines['CMAKE_CXX_FLAGS']} #{ENV['CXXFLAGS']}"
        if cxx_flags !~ /-std=c\+\+11/
            puts "cmake flags before: #{cxx_flags}"
            pkg.define "CMAKE_CXX_FLAGS", "#{cxx_flags} -std=c++11"
            puts "cmake flags after: #{pkg.defines['CMAKE_CXX_FLAGS']}"
        end
    else
        cxx_flags = ENV['CXXFLAGS']
        if cxx_flags !~ /-std=c\+\+11/
            puts "cxxflags before: #{cxx_flags}"
            ENV['CXXFLAGS'] = "#{cxx_flags} -std=c++11"
            puts "cxxflags after: #{ENV['CXXFLAGS']}"
        end
    end
end 
```

doudou:
1. do not iterate over Autobuild::Package. Autoproj.post_import is the way to go
2. Setting ENV on autoproj v1 will set CXXFLAGS globally. If it worked for you, it was by chance. On autoproj v2, we could use pkg.env_set which sets the flags locally to a particular package's build, but that would work only if people install e.g. gems or packages through autoproj EVER (it's part of the environment, so it will be visible as soon as an env.sh gets loaded)

In summary: don't try to change the choices of other package maintainers, it will only backfire on the rock developers. Change the C++11 setting only on Rock packages. _In fine_, ubuntu and others will switch one day to C++11/C++14/C++17 by default and that will be the end of it.

We *will* have to live in a mixed C++98/C++11 environment in any case since Ubuntu builds its packages with C++98 by default.

doudou:
> We will have to live in a mixed C++98/C++11 environment in any case since Ubuntu builds its packages with C++98 by default.

To explain why this is important: building non-rock packages forcefully with C++11 is only moving the C++98/C++11 interface, but does not change its surface. So, best leave other people's code alone (compiled with C++98) IMO.

doudou:
Superseded by #95

PR #89, opened 2016-07-18
Fix websocket version to <=1.2.2

2maz:
see rock-core/tools-syskit#101


doudou:
since the problem seem to be only with Ruby 2.0 (we've been using 2.1+ and did not have problems), I would prefer putting it in the osdeps specific to ruby2.0

2maz:
Sure  - updated the merge request accordingly

2maz:
ping

PR #91, opened 2016-08-11
Fix gems to specific versions to keep rack compatible with ruby2.0

jmachowinski:


jmachowinski:
Tested, fixes the thin / rake needs ruby 2.3 issue for me

doudou:
Can you *please* stop merging your own pull requests 10 minutes after you create them ? This only demonstrates how little you understand the very point of them.

PR #93, opened 2016-08-12
modify to be compatible with both autoproj v1 and v2

doudou:
some APIs changed in autoproj v2 (that's why it's a major version bump after all). This makes sure that the flavor code is compatible both with the old and new APIs.


doudou:
Merging it will demand that we synchronize with the release of the next 2.0.rc in which the APIs do change.

doudou:
Ping

doudou:
It's been a month ... fuck it.

PR #99, opened 2016-08-30
Set CMake flags in post_import phase

chhtz:
It seems that `pkg.tags` is not set at the moment `Rock.update_cmake_build_type_from_tags(pkg)` is called.
I reverted this part to a cleaned up version before https://github.com/rock-core/package_set/pull/92

### DIFF
--- a/overrides.rb
+++ b/overrides.rb
@@ -70,7 +70,9 @@ Autoproj.manifest.each_autobuild_package do |pkg|
             pkg.optional_dependencies.delete 'ocl'
         end
     when Autobuild::CMake
-        Rock.update_cmake_build_type_from_tags(pkg)
+        pkg.post_import do
+            Rock.update_cmake_build_type_from_tags(pkg)
+        end
         pkg.define "ROCK_TEST_ENABLED", pkg.test_utility.enabled?
         pkg.define "CMAKE_EXPORT_COMPILE_COMMANDS", "ON"
     end


jmachowinski:
I can confirm this, only works if set in post import

doudou:
@chhtz: are you running autoproj v1 or v2 ? Could you try defining it inside a setup_package block instead ?, e.g.

```
   when Autobuild::CMake
        setup_package(pkg) do
           Rock.update_cmake_build_type_from_tags(pkg)
        end
        pkg.define "ROCK_TEST_ENABLED", pkg.test_utility.enabled?
```

Note also that the defines should stay where they are unless there also is a problem with them in that block.

chhtz:
I'm on autoproj v1.13.5.
With the `setup_package` I get: 
```
#<Autobuild::CMake:0x000000013a4458> is not a known package
  in /home/chtz/workspace/entern/autoproj/remotes/rock.core/overrides.rb:73:in `block in <top (required)>'
  in /home/chtz/workspace/entern/autoproj/remotes/rock.core/overrides.rb:35:in `<top (required)>'
```
Removing the `setup_package` (i.e., reverting to without my PR) sets all my packages to `DEBUG`

doudou:
Sorry, on v1 it should be `setup_package(pkg.name) do`

chhtz:
With that it switches everything to Debug again:
```
$ amake base/types/
[...]
autoproj: importing and loading selected packages
  base/types: changed value of CMAKE_BUILD_TYPE from Release to Debug
  base/cmake: changed value of CMAKE_BUILD_TYPE from Release to Debug
[...]
```

chhtz:
N.B.: I'm not sure about the `ROCK_TEST_ENABLED`, `CMAKE_EXPORT_COMPILE_COMMANDS`
At the moment it seems to work in either place (I thought I had an issue with one of them as well).

doudou:
The issue is related to the order between loading of manifest.xml (the tags) and the call of setup_package. The defines are not dependent on files in the package, so are not subject to the same constraints. I just read the whole file, and it seems that it is the only one that has this problem.

Anyways, last one: could you try just doing

```
pkg.post_import do
   ... handle tags ...
end
```

at the original place, instead of moving it to a different place in the file ? It would help readability. Thanks a lot for your help.

chhtz:
Seems to work as well. I overwrote the commit and only changed handling of the build type now.

doudou:
Mmm ... miscommunication.

Could you try leaving the tags stuff in the `case ... when ... end` block, but surrounding it with the (arguably better)

```
pkg.post_import do
    Rock.update_cmake_build_type_from_tags(pkg)
end
```

?

chhtz:
Sorry for the delay and the misunderstanding. I updated the PR (meaning this variant works as well).

PR #100, opened 2016-08-31
Fix spurious creation of package srcdir

doudou:


jmachowinski:
That fixes the checkout bug

PR #101, opened 2016-09-13
remove vtk-qt4 on 16.04

doudou:
Extensive discussion:
    https://github.com/rock-core/rock-package_set/pull/129

Basically, the only user of this osdep on rock.core is
rock_widget_collection, and its support is disabled by default
(e.g. enabled virtually nowhere). Given that this conflicts
on 16.04 with dependencies needed by PCL, a much more needed
package, let's just remove vtk-qt4 on 16.04.

### DIFF
--- a/rock.osdeps
+++ b/rock.osdeps
@@ -80,8 +80,8 @@ qwt5:
 
 vtk-qt4:
     ubuntu:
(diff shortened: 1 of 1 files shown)


jmachowinski:
I think we all agree on this, merged

PR #102, opened 2016-09-20
Fix rack to older version

martinfritsche:
The current version of rack needs ruby version >= 2.2.2. So users of ruby2.1 need an older version of rack. (change copied from rock.osdeps-ruby20)


doudou:
Duplicate of https://github.com/rock-core/package_set/pull/96

PR #103, opened 2016-10-24
removed nonexistring osdep autotools for 16.04 LTS

hemker:
I observed an odd behavior with autoproj2 and Ubuntu 16.04 bootstraps. In 16.04 release there is no corresponding osdep "autotools" anymore. The interesting thing is, that if one requires the osdep rice -> autotools osdep chain in multiple *.osdeps sources and it's blocked in one file with '#' lines, the error "package autotools is not available as an osdep" does not occur. But if the only source with rice -> autotools is from rock.core, autoproj update when bootstrapping is not possible with autoproj2 and Ubuntu 16.04

### DIFF
--- a/rock.osdeps
+++ b/rock.osdeps
@@ -69,7 +69,9 @@ rice:
     default:
         gem: rice
         # Note: autoconf is an osdep built-in autoproj
-        osdep: autotools
+        osdep:       
+            default: nonexistent
+            '14.04,14.10,15.04,15.10': autotools
 
 qwt5:
     debian,ubuntu: libqwt5-qt4-dev


jmachowinski:
I don't get it, were is autotools defined in the first place ?!
I did a grep in the remotes but can't find a definition anywhere.

hemker:
It's primarily defined in rock.core/rock.osdeps. It's overriden in - for instance - cave/cave.osdeps, so in the entern bootstrap one won't get the error

jmachowinski:
No, it is set as a 'dependency' of rice.
But it is not defined there. There is no place, were is
written, which apt package should be installed.

hemker:
Primary it's defined rock.core.osdeps and it's overriden - for instance - im the cave osdeps 

Am 26. Oktober 2016 09:55:54 MESZ, schrieb jmachowinski <notifications@github.com>:

doudou:
Autotools is a built-in osdep from autoproj itself. Any autotools_package depends on it, the same way that any cmake_package depends on the cmake osdep

See [default.osdeps](https://github.com/rock-core/autoproj/blob/master/lib/autoproj/default.osdeps)

doudou:
>  The interesting thing is, that if one requires the osdep rice -> autotools osdep chain in multiple *.osdeps sources and it's blocked in one file with '#' lines, the error "package autotools is not available as an osdep" does not occu

I don't understand what you mean one bit ...

doudou:
In any case, this is a wrong patch. Autotools did not disappear from 16.04, the metapackage called 'autotools' never existed in the first place. If you look at autoproj's default.osdeps, you'll see that autools basically installs automake and autoconf.

Could it be that this 'cave' package set of yours is messing up ?

hemker:
in rock.core, following is inserted:
```rice:
default:
    gem: rice
    # Note: autoconf is an osdep built-in autoproj
    osdep: autotools
```

Using the spacebot bootstrap, executing an autoproj update causes following error:
          
```
Command failed  
osdep autotools does not exist. It is referred to by rice.: Autoproj::OSPackageResolver::InvalidRecursiveStatement
```

doudou:
Which version of autoproj are you running ?

You can find out with `bundle show autoproj`

hemker:
I guess the latest:
`.autoproj/gems/ruby/2.3.0/gems/autoproj-2.0.0.rc40`

doudou:
Indeed, but it's always worth verifying ;-)

In any case ... I don't dispute that you experience the problem. However, I can't reproduce it (rice installation works like a charm on all the 16.04s of my team), and what you propose as a change makes no sense. 

I don't even understand the change. 'nonexistent' will *explicitely* make autoproj fail on rice on 16.04. What gives ?

hemker:
removing `osdep: autotools` makes autoproj update work again for my case spacebot

doudou:
Something is puzzling me ...

You wrote:

> > I don't get it, were is autotools defined in the first place ?!
> It's overriden in - for instance - cave/cave.osdeps, so in the entern bootstrap one won't get the error

But then you say

> autotools as dependency is not mentioned in any other package set from spacebot

?

hemker:
Alright, i mean:

When I checkout the entern stack, `rice` is is referred in two package sets - cave and rock.core. Cave overrides the rice entry in cave.osdeps as following:
```
rice:
    default:
        gem: rice
        # Note: autoconf is an osdep built-in autoproj
        # osdep: autotools 
```
So the original dependency of `autotools` as an osdep from rock.core gets overriden.
Conclusion: No problem when using entern softwarestack.

In the spacebot softwarestack, `rice` is nowhere referred but solely in rock.core/rock.osdeps. And the entry looks as following:

```
rice:
    default:
        gem: rice
        # Note: autoconf is an osdep built-in autoproj
        osdep: autotools 
```
And the dependency `autotools` as an osdep is *not* overriden anywhere.
Conclusion: the autoproj error `Command failed  
osdep autotools does not exist. It is referred to by rice.: Autoproj::OSPackageResolver::InvalidRecursiveStatement`

I hope it's somehow clearer now

doudou:
> I hope it's somehow clearer now

I get it now ... but still don't get the bug (or, more accurately, why you get the bug and I don't)

I just release rc41. 'show' should be more informative about the osdeps. Could you update with

    autoproj update --autoproj

and then run

    autoproj show autotools

hemker:
Following your instructions, the entern installation produces following output:
```
the osdep 'autotools'
  apt-dpkg: automake, autoconf, build-essential
  1 matching entries:
    in /home/hemker/.autoproj/gems/ruby/2.3.0/gems/autoproj-2.0.0.rc41/lib/autoproj/default.osdeps:
        gentoo:
      - sys-devel/automake
      - sys-devel/autoconf
      default:
      - automake
      - autoconf
      osdep: build-essential
  is not directly selected by the manifest
  direct reverse dependencies: drivers/aravis, 
``` 
and lots of other packages after aravis.

On spacebot installation, the command failes with the error frequently mentioned above. Uncommenting the autotools osdep produces:
```
the osdep 'autotools'
  apt-dpkg: automake, autoconf, build-essential
  1 matching entries:
    in /home/hemker/.autoproj/gems/ruby/2.3.0/gems/autoproj-2.0.0.rc41/lib/autoproj/default.osdeps:
        gentoo:
      - sys-devel/automake
      - sys-devel/autoconf
      default:
      - automake
      - autoconf
      osdep: build-essential
  is not directly selected by the manifest
  direct reverse dependencies: drivers/aravis
```
again followed by lots of packages

doudou:
> Uncommenting the autotools osdep produces:

You mean "commenting the osdep: autotools line in the rice definition", right ?

hemker:
Yep

doudou:
At this stage ... apart from a good'ol debugging session, I don't see how to move forward. Could you tarball your autoproj/ and .autoproj/remotes/ directories ? I could then at least try to see if I can reproduce it here.

hemker:
Here we go

hemker:
I send you a mail with the tarball because it's not a supported upload extension here

doudou:
And there we are. The bug is within the piece of code that attemps to check whether an osdep is overriden (and warn about it). 'rice' is overriden in rock.dfki. Removing that override un-triggers the bug.

hemker:
Thanks, so I guess this PR is obsolete now.

PR #107, opened 2016-11-30
Orogen packages should also go through the generic CMake package setup

doudou:


rafaelsaback:
This change has solved the issue of not setting `CMAKE_BUILD_TYPE` to Debug when it was expected to be set. :+1:

PR #109, opened 2017-01-27
switch to git repo for libsisl

planthaber:
pending pull request on the git repo to build with gcc6

### DIFF
--- a/source-stable.yml
+++ b/source-stable.yml
@@ -6,9 +6,8 @@ version_control:
       url: https://bitbucket.org/eigen/eigen/get/3.2.1.tar.gz
       update_cached_file: false
     - external/sisl:
-      type: archive
-      url: http://www.sintef.no/upload/IKT/9011/geometri/sisl/sisl-4.5.0.tar.gz
-      update_cached_file: false
+      github: SINTEF-Geometry/SISL
+      tag: SISL_4_5_0
     - external/opencv:
       type: archive
       url: http://downloads.sourceforge.net/project/opencvlibrary/opencv-unix/2.3.1/OpenCV-2.3.1a.tar.bz2
--- a/source.yml
+++ b/source.yml
@@ -66,9 +66,8 @@ version_control:
       branch: $ROCK_BRANCH
 
     - external/sisl:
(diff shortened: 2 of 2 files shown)


jmachowinski:
the pull request was just merged

planthaber:
saw that, now we have to switch to actually use the git repo instaed of the tar

jmachowinski:
Write an mail about this to the rock-users ml, and I'll merge

planthaber:
Ah yes, this will cause the "external/sisl" is not a git repository error...
I'll write the mail.

doudou:
Write email and then *3 hours later* merge. For a change that will make everyone's next update fail.

doudou:
Next time you'll ask about waiting for publishing say something like autoproj 2, I'll consider how *you* handle everybody else's installation.

planthaber:
Except this is a bugfix which makes rock build again on 16.10 and debian testing:
https://github.com/SINTEF-Geometry/SISL/pull/9

doudou:
> Except this is a bugfix which makes rock build again on 16.10 and debian testing:

Do realize that I don't object to the change - I think it's great - but to the timing of it.

To have it working on your pet ubuntu you're ready to break everyones updates with no advance warning ... I knew that we definitely don't have the same priorities, this only confirms it.

planthaber:
I am using the current Ubuntu because we promised rock would run on Ubuntu LTS and current version. I not particularly like using the current version and to be the one to adapt osdeps and solve bugs like these and gcc6 stuff and all the other issues. But someone has to to it.

Otherwise we should discuss to supporting only LTS Versions on the mailing list.

Time between mail and merge could have been larger, right. But that is a discussion between those who are able to merge, and i am not able to merge stuff within rock-core at all

jmachowinski:
I don't see how this change can brick an installation. It's the same version of SISL with one additional patch. Apart from this the user needs to once delete his sisl folder and reupdate. 
The patch was checked beforehand on ubuntu 14.04 16.04 and 16.10.
Rebuilding a stack takes around 5 minutes, as everything is in ccache...

PR #113, opened 2017-05-25
allow forcing mix between the master flavor and the presence of a release file

doudou:
Running 'autoproj status' is a necessary step during the creation
of a release candidate / during a release. Allow bypassing this
check when preparing a release by setting ROCK_RC to 1

### DIFF
--- a/init.rb
+++ b/init.rb
@@ -51,11 +51,19 @@ current_flavor = Rock.flavors.current_flavor
 #This check is needed because the overrides file will override the FLAVOR selection.
 #Furthermore a selection != stable can cause a inconsistent layout (cause by in_flavor system in the package_sets)
 if File.exists?(File.join(Autoproj.root_dir, "autoproj", "overrides.d", "25-release.yml")) && current_flavor.branch != "stable" 
-    Autoproj.error ""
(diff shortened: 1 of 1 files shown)


init.rb
```diff
@@ -50,7 +50,7 @@
 
 #This check is needed because the overrides file will override the FLAVOR selection.
 #Furthermore a selection != stable can cause a inconsistent layout (cause by in_flavor system in the package_sets)
-if File.exists?(File.join(Autoproj.root_dir, "autoproj", "overrides.d", "25-release.yml")) && current_flavor.branch != "stable" 
+if ENV['ROCK_RC'] != '1' && File.exists?(File.join(Autoproj.root_dir, "autoproj", "overrides.d", "25-release.yml")) && current_flavor.branch != "stable"
```
2maz: I would suggest to increase verbosity, i.e. something along the following lines:
```
if ENV['ROCK_RC'] == 1
    Autoproj.warn "You are operating on a Rock Release Candidate"
elsif ...
end

doudou: :+1:. I've done something that looks like that. Could you check whether the message makes sense to you ?

PR #115, opened 2017-05-30
fix typelib_cxx_loader override logic

doudou:
Hopefully fixes https://github.com/rock-core/package_set/pull/114 that is hopelessly broken

This time with tests (which found more problems)

### DIFF
--- a/rock/cxx11.rb
+++ b/rock/cxx11.rb
@@ -1,25 +1,37 @@
 module Rock
+    def self.has_cxx11_support?(os_package_resolver)
+        os_package_resolver.has?('libclang-castxml') || os_package_resolver.has?('castxml')
+    end
+
+    def self.default_loader_is_castxml?(os_package_resolver)
+        os_package_resolver.has?('default_castxml')
+    end
+
     # Offer the C++11 option only on systems where we either can install castxml
     # from the OS packages, or build it.
     #
     # But then, set it to true on OSes where it should. This is controlled by
     # stub osdep entries in defaults.osdeps
     def self.setup_cxx11_support(os_package_resolver, config)
-        if os_package_resolver.has?('libclang-castxml') || os_package_resolver.has?('castxml')
+        has_cxx11_support = has_cxx11_support?(os_package_resolver)
+        if has_cxx11_support
             config.declare 'cxx11', 'boolean',
                 default: 'no',
                 doc: "whether C++11 should be enabled for Rock packages"
+            cxx11_enabled = config.get('cxx11')
         else
-            config.set 'typelib_cxx_loader', 'gccxml'
             config.set 'cxx11', false
+            cxx11_enabled = false
         end
 
-        if !config.has_value_for?('typelib_cxx_loader') && os_package_resolver.has?('default_castxml')
-            config.set 'typelib_cxx_loader', 'castxml', true
-        elsif config.get('cxx11')
-            config.set 'typelib_cxx_loader', 'castxml'
-        else
-            config.set 'typelib_cxx_loader', 'gccxml'
+        if !config.has_value_for?('typelib_cxx_loader')
+            if cxx11_enabled
+                config.set 'typelib_cxx_loader', 'castxml'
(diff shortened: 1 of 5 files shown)


test/cxx11_test.rb
```diff
@@ -0,0 +1,90 @@
+require 'test_helpers'
```
2maz: I had to use require_relative to get the tests running

test/cxx11_test.rb
```diff
@@ -0,0 +1,90 @@
...
+                assert_equal 'something', @config.get('typelib_cxx_loader')
+            end
+
+            it "sets the loader to castxml if it is the default on this OS" do
+                Rock.should_receive(:default_loader_is_castxml?).and_return(true)
+                Rock.setup_cxx11_support(@os_package_resolver, @config)
+                assert_equal 'castxml', @config.get('typelib_cxx_loader')
+            end
+
+            it "sets the loader to castxml event if it is not the default on this OS" do
```
2maz: event -> even

2maz:
Tests pass, and looks good on my test bootstrap (xenial/autoproj v2).
However, I would expect gccxml + cxx11 also to work for trusty, but cxx11 enabled is set to false after an 'autoproj reconfigure'.

test/cxx11_test.rb
doudou: You have to run with `-I. -Itest`, or just "rake test"

doudou:
> However, I would expect gccxml + cxx11 also to work for trusty, but cxx11 enabled is set to false after an 'autoproj reconfigure'.

This was already the previous behavior. There is a check whether clang/castxml is available and if it's not it just does not even ask about C++11.

What do you think should be the behavior ?

test/cxx11_test.rb
doudou: I've added a README.md to explain how to run the tests

2maz:
> This was already the previous behavior. There is a check whether clang/castxml is available and if it's not it just does not even ask about C++11.
Yet, it is not clear why checking for cxx11 support actually redirects to only checking for castxml support . 
So, since I might miss the obvious, why is cxx11: false enforced when castxml is not available?

doudou:
> So, since I might miss the obvious, why is cxx11: false enforced when castxml is not available?

I guess that my idea was to avoid C++11 if orogen/typelib had no chance to handle it (which is a bit of a moot point given that C++11 parsing is still not officially available in orogen). I wouldn't mind removing that check though.

2maz:
Ok, but even then the actual types that orogen need to handle have to be compatible (i.e. don't use C++11 features) while the libraries can still be compiled with cxx11?!
I would suggest to remove that check for now.

doudou:
> Ok, but even then the actual types that orogen need to handle have to be compatible (i.e. don't use C++11 features) while the libraries can still be compiled with cxx11?!

Yes. That was my point when I said

> which is a bit of a moot point given that C++11 parsing is still not officially available in orogen

> I would suggest to remove that check for now.

OK. Will do

doudou:
Updated

rock/cxx11.rb
```diff
@@ -1,8 +1,4 @@
 module Rock
-    def self.has_cxx11_support?(os_package_resolver)
```
2maz: --> has_castxml_support?

rock/cxx11.rb
```diff
@@ -13,21 +9,15 @@ def self.default_loader_is_castxml?(os_package_resolver)
...
-            config.set 'cxx11', false
-            cxx11_enabled = false
-        end
+        config.declare 'cxx11', 'boolean',
+            default: 'no',
+            doc: "whether C++11 should be enabled for Rock packages"
+        cxx11_enabled = config.get('cxx11')
 
         if !config.has_value_for?('typelib_cxx_loader')
             if cxx11_enabled
```
2maz: has_castxml_support?

doudou: Since there's already the castxml-is-default test, I just went for that.

2maz:
Ok, merged.

PR #117, opened 2017-06-05
move iodrivers_base from rock to rock-core

doudou:


jmachowinski:
I don't get the purpose of the patch, should it delete the test folder in the build dir ?
Why ?

doudou:
> I don't get the purpose of the patch, should it delete the test folder in the build dir ?
Why ?

The patch simply copies the package definition from the rock package set. The post_import block was there because at some point there was a 'test' file in the build directory, which conflicted with the proper test folder (and updates would fail).

jmachowinski:
Okay, 
what about house keeping and remove the block about the test directory ?

doudou:
> what about house keeping and remove the block about the test directory ?

I could even enjoy interacting with you if, sometimes, you would try to do more than just  say things. Like, actually show that you considered the implications and did the little bit of research that would be actually helpful. Or, you know, actually review pull requests before you merge them. Write tests for bugfixes in pull requests. The "housekeeping" that would make rock core better little by little.

Yes, I did think about it. Now, given that I'm actually spending the effort of doing all the "housekeeping" of making a release, I passed it on because it is inconsequential.

"Oh, I looked, and this block was introduced in 2011, way before the last release. We should probably remove it now" would have seen a much better reception.

doudou:
Since the repo has been modified to match the there was ... I merge.

jmachowinski:
I can't merge any more, you revoked my maintainer status....

doudou:
> you revoked my maintainer status....

By accident. I revoked your release manager status, under the assumption that you would still have maintainer rights through your Core team membership.

doudou:
OK... Here's the situation.

Per https://rock.opendfki.de/trac/wiki/WikiStart/Standards/RG9 there's a bunch of packages whose PRs can be merged only by the org owners. It happens that you've never been one, but started having that privilege because you've been a release manager.

So, you lost privilege on the "really core" packages as defined in the above link.

But, again according to this process, the whole point of this restriction is to reduce the amount of mistakes made on these critical packages. *Not* give a bunch of people (namely, Thomas and me) the right to push whatever they want. A.k.a. we still expect other core developers to say "OK to merge" before we merge.

jmachowinski:
what is you point, that you just broke this rule ?

doudou:
> that you just broke this rule ?

It does seem so, yes.

PR #120, opened 2017-06-28
run the setup_cxx11_support within init.rb

doudou:
Otherwise, it is run after the package setup blocks, whose behavior
obviously might depend on the result.

Since osdeps files are not yet loaded in init.rb, load the default
files explicitly there, and avoid polluting the osdep definitions
with it (since the defaults files uses the osdep system without
actually defining osdeps)


planthaber:
Just tested it, it fixes the issue.

So, please merge.

Thanks

PR #122, opened 2017-07-13
fix castxml/gccxml selection under autoproj v1

### LINKED ISSUE #121: cxx11.rb incompatible with autoproj v1
cxx11.rb: Autoproj::OSPackageResolver is only known to autoproj v2, and effectively preventing the continued use of autoproj v1 installations without upgrading

doudou:
Fixes #121


Rauldg:
Fixes the issue. 

Without this fix autoproj v1 failed at checking out the rock core in my case.

PR #125, opened 2017-08-02
Add the dependency of the GTest framework

alcantara09:
With https://github.com/rock-core/drivers-iodrivers_base/pull/20, iodrivers_base test suite would become dependent on GTest (only for the tests). This moves the gtest osdeps from rock to rock.core.

### DIFF
--- a/rock.osdeps
+++ b/rock.osdeps
@@ -224,8 +224,13 @@ rgl: gem
 autorespawn: gem
 timecop: gem
 
-libgtest-dev:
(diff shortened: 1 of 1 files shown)


doudou:
Btw, this has conflicts with master, please rebase and fix them first.

doudou:
:+1: I've only renamed libgtest-dev into google-test for consistency, and added libgtest-dev as an alias for backward compatibility

PR #126, opened 2017-08-07
Fix setting of ROCK_WEBAPP

2maz:


00new_packages.autobuild
```diff
@@ -13,5 +13,5 @@ end
 
 cmake_package 'gui/rock_webapp'
 ruby_package 'tools/rest_api' do |pkg|
-    Autoproj.env_set 'ROCK_WEBAPP', File.join(pkg.prefix, 'share', 'rest_api')
+    Autoproj.env_set 'ROCK_WEBAPP', File.join(Autoproj.build_dir, 'share', 'rest_api')
```
doudou: Why this change ? The env var should not refer to the global build dir / prefix.

2maz: Where else would the "share" folder be? I could not find any in tools/rest_api?
So that change was required to make the rest_api actually work for us (with autoproj v2) since otherwise the plugins are not found.

Why should an env var not refer to the build dir? Sounds to me a bit arbitrary.

doudou: > Why should an env var not refer to the build dir ?

Because it assumes that all the package prefixes are the main autoproj build directory. Which is not true if you use autoproj's extremely useful `separate_prefixes` feature.

For a more documented use-case, one that would create a pure ruby rock-webapp plugin can't use it.

In any case, the proper fix would be to have ROCK_WEBAPP a PATH instead of a single-path envvar. In the meantime, it could be set by gui/rock_webapp to wherever it is installed, instead of assuming a global install location.

planthaber: ROCK_WEBAPP defines the configuration directory so a single path variable is wanted.
Inside the configuration directory a "plugins" folder may esist, which is evaluated

I think you should use Autobuild.prefix instead of Autoproj.build_dir to fix this this config dir issue.

For the plugins one could think about adding a additional ROCK_WEBAPP_PLUGIN_PATH but i remember vaguely that for some reason we decided to do an "install" for the plugins.

Having this new plugin path would be fine for me. but this is unrelated to the faulty setting of the config dir


Autobuild.prefix

doudou: If you don't have the time to change ROCK_WEBAPP into a path, then set the envvar into `gui/rock_webapp`  to `pkg.prefix`. And everyone should do the same if they expect to export a webapp "configuration directory".

It will work in cases where everyone is installed in the same prefix.

It will make autoproj fail if two packages try to set ROCK_WEBAPP in two different places.

planthaber: * ROCK_WEBAPP is a single path solely used as configuration directory for the rock-rest-api program, this is a singeton by definition, as there cannot be two configurations for a single instance of a program.
* the envvar WAS already pkg.prefix, but with autoproj 2 this points to the SOURCE directory for ruby packages, not the installation directory (that is why this PR was created)

Apart from the different behavior between autoproj 1 and 2 concerning pkg.prefix for ruby packages, we could indeed think about adding the mentioned ROCK_WEBAPP_PLUGIN_PATH, which then, should be a real path with multiple entries.

But adding that is a new feature to the REST API, while this PR is a simple bugfix (or workaround to a autoproj bug).

doudou: If you would read my comment more carefully, you would notice that I say that the setting should be set for **gui/rock_webapp**'s prefix, not **tools/rest_api**.

The very fact that taking the prefix of `tools/rest_api` does not work should have rang a bell already. It's not working because **it's not the right prefix**.

PR #127, opened 2017-08-10
fix webapp path

planthaber:
RubyPackage#prefix  changed from pointing to install folder to package location in autoproj 2

So now this uses the global install folder. 

This is also more "correct", because this is a plugin path which is populated by other packages. I think it should not rely on the prefix settings of the rest api package.


doudou:
Duplicate of #126

PR #131, opened 2018-02-20
use the 0.3.8 release for class_loader

doudou:
- the tip of indigo_devel does not build right now
- 0.3.8 is the last release of the 0.3.x branch, which is
  what we've been using so far
- 1.0.0 has been released, but I'd rather propose a PR that switches to 1.0.0 later,
  to give time for testing.


dbcesar:
Fine for me. I started discussion there, but it can take some time until is finishes.

PR #132, opened 2018-03-07
update templates env vars only if cmake_lib package is imported

g-arjones:


doudou:
Just use the package-level environment setters instead of the global ones, e.g. `pkg.env_set`.

We need to review the whole package set to convert everything to those.

g-arjones:
Cool. Updated :+1:

PR #134, opened 2018-05-04
add xmlrpc gem dependency (for Ubuntu 18.04)

planthaber:


rock.osdeps
```diff
@@ -247,3 +247,10 @@ rest-client:
 
 ruby-backports:
     gem: backports
+
+xmlrpc:
```
doudou: It's a ruby version issue, not a Ubuntu version issue. I.e. one would have the same problem with Ruby 2.5 installed with rbenv on a Ubuntu 17.10. Since we support only Ruby >= 2.3 and the xmlrpc gem is compatible with ruby >= 2.3, just install the gem.

Moreover, given how generic the 'xmlrpc' name is, I would suggest renaming the osdeps to `ruby-xmlrpc`.

planthaber: force-pushed both changes

PR #135, opened 2018-05-09
added osgviz package

planthaber:
As discussed in https://github.com/rock-core/gui-vizkit3d/pull/34


doudou:
Wouldn't it make more sense to integrate the various osgviz packages all at once ? (I'm really asking, I have no idea how the dependencies / functionality look like)

planthaber:
We didn't release them to opensource jet, but as they are more "beta" and also not really needed in the core, i'd leave them defined here:

https://github.com/rock-gui/osgviz-package_set

PR #136, opened 2018-10-09
Add apaka: Automated (Debian) packaging for autoproj

2maz:


doudou:
Should we remove it from the default install, the way admin_scripts is ?

2maz:
Yap, updated accordingly.

PR #138, opened 2018-10-26
Add tools/pocolog2msgpack

AlexanderFabisch:
@2maz Could you check if I specified everything correctly? Do I also have to specify the optional Python packages?

### DIFF
--- a/rock.osdeps
+++ b/rock.osdeps
@@ -251,3 +251,18 @@ ruby-backports:
 ruby-xmlrpc:
    gem: xmlrpc
 
+python:
+    debian,ubuntu: python
+    opensuse: python
+
+python-nose:
+    debian,ubuntu: python-nose
+    opensuse: python-nose
+
+python-pexpect:
+    debian,ubuntu: python-pexpect
+    opensuse: python-pexpect
+
+python-msgpack:
+    debian,ubuntu: python-msgpack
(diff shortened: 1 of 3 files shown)


2maz:
If the python packages are not yet defined in an osdeps file, then pls add them.
Otherwise specification looks ok.

rock.osdeps
```diff
@@ -251,3 +251,18 @@ ruby-backports:
 ruby-xmlrpc:
    gem: xmlrpc
 
+python:
```
AlexanderFabisch: Not sure if Python has to be defined or whether it is already known to autoproj.

2maz: Currently it is defined in rock package_set. We can remove it there, once the osdeps has been defined in rock.core.

rock/rock.osdeps
```
python:
    debian, ubuntu: python-dev
    opensuse: python-devel
```

AlexanderFabisch: OK, should be fixed now...

AlexanderFabisch:
Yep, here: https://github.com/rock-core/rock-package_set/pull/127

2maz:
Ok, I think you meant rock-core/rock-package_set#182
Thanks, anyhow.

AlexanderFabisch:
Yes, I don't know how that happened. 😀

PR #144, opened 2019-08-29
move setup of test utility to post_import

doudou:
It turns out that setting up the test utility requires a checkout
(we check for the existence of a test/ folder). Autoproj is moving
the autodetection to within a post_import context.

Autoproj fix will keep backward compatibility (the setup is done twice, once at setup time
and once in the post_import context), but best to move the setup on our side to the right
place right away.

Regression test: https://github.com/rock-core/build_tests-test/pull/4

### DIFF
--- a/overrides.rb
+++ b/overrides.rb
@@ -75,9 +75,11 @@ Autoproj.manifest.each_autobuild_package do |pkg|
     end
 
     if pkg.kind_of?(Autobuild::Ruby)
-        if pkg.test_utility.enabled? && pkg.respond_to?(:rake_test_options) # autoproj v2 only
-            pkg.depends_on 'minitest-junit'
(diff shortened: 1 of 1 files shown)


overrides.rb
```diff
@@ -75,9 +75,11 @@
     end
 
     if pkg.kind_of?(Autobuild::Ruby)
-        if pkg.test_utility.enabled? && pkg.respond_to?(:rake_test_options) # autoproj v2 only
-            pkg.depends_on 'minitest-junit'
-            pkg.rake_test_options << "TESTOPTS=--junit --junit-filename=#{pkg.test_utility.source_dir}/report.junit.xml --junit-jenkins"
+        pkg.post_import do
```
g-arjones: post import hooks are called even when don't do an import operation, right?

doudou: Yes. From the perspective of Autoproj 'import' is 'possibly import and load configuration'

PR #148, opened 2019-11-11
pin rtt commit to before the merge of the last major dataflow refactoring

doudou:
This refactoring led to massive deadlocking on a lot of dynamic
disconnection scenarios.

See https://github.com/orocos-toolchain/rtt/pull/302

### DIFF
--- a/source.yml
+++ b/source.yml
@@ -91,6 +91,15 @@ version_control:
 overrides:
     - ^(orogen|typelib|rtt|utilrb|utilmm|rtt_typelib|tools/metaruby)$:
       branch: $ROCK_BRANCH
+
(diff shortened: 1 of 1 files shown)


source.yml
```diff
@@ -91,6 +91,8 @@ version_control:
 overrides:
     - ^(orogen|typelib|rtt|utilrb|utilmm|rtt_typelib|tools/metaruby)$:
       branch: $ROCK_BRANCH
+    - rtt:
```
2maz: Could you add a comment that this is temporary fix with link to the PR / reason why this is needed.
Otherwise approved from my side.

doudou: Very good point ... Done.

PR #149, opened 2019-11-11
pull the value of ROCK_TEST_BOOST_FORMAT from configuration

doudou:
This is a value that typically needs to be set in CI environments,
make sure that we can do so easily


overrides.rb
```diff
@@ -93,6 +93,8 @@
             pkg.define 'ROCK_TEST_LOG_DIR', pkg.test_utility.source_dir
         end
         pkg.define "CMAKE_EXPORT_COMPILE_COMMANDS", "ON"
+        pkg.define 'ROCK_TEST_BOOST_FORMAT',
+                   pkg.ws.config.get('rock_test_boost_format', 'XML')
```
g-arjones: I think this should be defined only when tests are enabled... CMake will generate warnings for unused definitions

doudou: Done

PR #155, opened 2020-01-21
Simplify setting of python path for packages with bindings

2maz:
Example:

cmake_package 'foo' do |pkg|
    Rock.activate_python(pkg)
end

### DIFF
--- a/rock/python.rb
+++ b/rock/python.rb
@@ -1,0 +1,201 @@
+require 'open3'
+require 'rubygems'
+# Usage:
+# In init.rb
+#      Rock.activate_python()
+#
+# In lib.autobuild
+#      cmake_package 'tools/yourpackage' do |pkg|
+#          Rock.activate_python_path(pkg)
+#      end
+#
+# Assumptions:
+#     0. a python installation must be available
+#     1. only a single python version is set in the configuration for all
+#           packages  (vs. storing a configuration per package)
+#     2. activation takes place through a single call to Rock.activate_python in
+#               init.rb
+#     3. per package the python path can be updated/activated based on
+#              the existing configuration (default), or a custom binary
+#              (however, no caching is done for that part)
+#     4. version constraints are checked upon activation
+module Rock
+    # Get the python version for a given python executable
+    # @return [String] The python version as <major>.<minor>
+    def self.get_python_version(python_bin)
+        if !File.exist?(python_bin)
+            raise ArgumentError, "Rock.get_python_version executable "\
+                        "'#{python_bin}' does not exist"
+        end
+
+        cmd = "#{python_bin} -c \"import sys;"\
+            "version=sys.version_info[:3]; "\
+            "print('{0}.{1}'.format(*version))\"".strip()
+
+        msg, status = Open3.capture2e(cmd)
+        if status.success?
+            python_version = msg.strip()
+            return python_version
+        else
+            raise RuntimeError, "Rock.get_python_version identification"\
+                " of python version for '#{python_bin}' failed: #{msg}"
+        end
+    end
+
+    def self.validate_version(version, version_constraint)
+        if !version_constraint
+            return true
+        else
+            dependency = Gem::Dependency.new("python", version_constraint)
+            return dependency.match?("python", version)
+        end
+    end
+
+    # Validate that a given python executable's version fulfills
+    # a given version constraint
+    # @param [String] python_bin the python executable
+    # @param [String] version_constraint version constraint, e.g., <3.8, >= 3.7, 3.6
+    # @return [String,Bool] Version and validation result, i.e., True if binary fulfills the version constraint, false
+    #   otherwise
+    def self.validate_python_version(python_bin, version_constraint)
+        version = get_python_version(python_bin)
+        return [version, validate_version(version, version_constraint)]
+    end
+
+    # Find python given a version constraint
+    # @return [String,String] path to python executable and python version
+    def self.find_python(ws: Autoproj.workspace,
+                         version: ws.config.get('python_version',nil))
+        finders = [
+            lambda { Autobuild.programs['python'] },
+            lambda { `which python`.strip() }
+        ]
+
+        finders.each do |finder|
+            python_bin = finder.call
+            if python_bin
+                python_version, valid = validate_python_version(python_bin, version)
+                if valid
+                    return python_bin, python_version
+                end
+            end
+        end
+        raise RuntimeError, "Rock.find_python_bin: failed to find python" \
+            " for version '#{version}'"
+    end
+
+    # Get information about the python executable from autoproj config,
+    # but ensure the version constraint matches
+    #
+    # @return [String, String] Return path and version if the constraints
+    #      are fulfilled [nil,nil] otherwise
+
+    def self.get_python_from_config(ws: Autoproj.workspace, version: nil)
+        config_bin = ws.config.get('python_executable', nil)
+        if config_bin
+            config_version = ws.config.get('python_version', nil)
+            config_version ||= get_python_version(config_bin)
+
+            # If a version constraint is given, ensure fulfillment
+            if validate_version(config_version, version)
+                return config_bin, config_version
+            else
+                raise RuntimeError, "python_executable in autoproj config with version '#{config_version}'"\
+                    " does not match version constraints '#{version}'"
+            end
+        end
+        [nil, nil]
+    end
+
+    def self.custom_resolve_python(ws: Autoproj.workspace,
+                                   bin: nil,
+                                   version: nil)
+        version, valid = validate_python_version(bin, version)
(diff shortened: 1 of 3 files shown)


rock/python.rb
```diff
@@ -0,0 +1,50 @@
+require 'open3'
+
+module Rock
+    # Identify the available python version, per default checks
+    # Autobuild.programs['python'] setting
+    # @param[String] python_bin Path to a custom python version
+    def self.find_python_bin(python_bin = nil)
+        if !python_bin
+            python_bin = Autobuild.programs['python']
```
doudou: You could store it in the configuration file, to be consistent with the way the Ruby stuff is done.

Also, I'd suggest to auto-cache the detected value there to avoid calling which repeatedly.

rock/python.rb
```diff
@@ -0,0 +1,50 @@
...
+    end
+
+    # Allow to update the PYTHONPATH for package, tries to guess the python
+    # binary from Autobuild.programs['python'] and system's default setting
+    # @param[Autobuild::Package] pkg
+    # @param[String] python_bin Path to a custom python version
+    def self.activate_python(pkg, python_bin = nil)
+        python_bin = Rock.find_python_bin(python_bin)
+        if python_bin
+            cmd = "#{python_bin} -c \"import sys;"\
```
doudou: Same here, best would be to cache the version. This is most probably rather expensive to run, and all the blocks are run at each autoproj load.

rock/python.rb
```diff
@@ -0,0 +1,50 @@
+require 'open3'
+
+module Rock
+    # Identify the available python version, per default checks
+    # Autobuild.programs['python'] setting
+    # @param[String] python_bin Path to a custom python version
+    def self.find_python_bin(python_bin = nil)
```
doudou: Yeah. This makes no sense. Either it finds the python bin or it does not, calling find_python_bin as an alias for File.exist? is really confusing.

rock/python.rb
```diff
@@ -0,0 +1,50 @@
...
+        end
+    end
+
+    # Allow to update the PYTHONPATH for package, tries to guess the python
+    # binary from Autobuild.programs['python'] and system's default setting
+    # @param[Autobuild::Package] pkg
+    # @param[String] python_bin Path to a custom python version
+    def self.activate_python(pkg, python_bin = nil)
+        python_bin = Rock.find_python_bin(python_bin)
+        if python_bin
```
doudou: Please return early.

Also: do you actually want to return nil ? I'd rather raise at this point.

rock/python.rb
2maz: "Finding" initial means to identify the user preference on the python bin, e.g., by checking custom path, Autoproj.program
There is, however, no guarantee that the binary is available unless it is checked explicitly.

2maz:
Refactored and added the use of the configuration.
Not entirely sure whether the initial Autoproj.config.load is required.

rock/python.rb
doudou: My point is that this is very confusing:

~~~
if path = Autobuild.programs['python']
   find_python(path)
end
~~~

You need two codepaths, one that checks if the file exists if it is in programs, one that calls `find_python_bin`. This is an unnecessary (and very confusing) overload. Just the size of the `if` smells really bad.

doudou:
> Not entirely sure whether the initial Autoproj.config.load is required.

It's not only not required, it's harmful. If one has changed configuration parameters before calling find_python_bin, they will be overriden.

rock/python.rb
```diff
@@ -0,0 +1,66 @@
...
+                if status.success?
+                    python_version = msg.strip()
+
+                    if !custom_python_bin
+                        Autoproj.config.set('python_executable', python_bin)
+                        Autoproj.config.set('python_version', python_version)
+                    end
+                    return python_bin, python_version
+                else
+                    Autproj.warn "#{__FILE__} identification of python version"\
```
doudou: Your tests don't go through this code path ... (typo on 'Autoproj')

rock/python.rb
```diff
@@ -0,0 +1,66 @@
+require 'open3'
+
+module Rock
+    # Identify the available python version, per default use the set values in
+    # Autoproj configuration for 'python_executable' and 'python_version',
+    # otherwise checks Autobuild.programs['python'] setting
+    # @param[String] custom_python_bin Path to a custom python version
+    # @returns[[String,String]] path to the python binary and version, throws if
+    #   python cannot be found
+    def self.find_python_bin(custom_python_bin = nil)
```
doudou: I'm trying to get rid of globals access as much as possible. A better pattern would be to pass a `ws` keyword argument (an Autoproj::Workspace object, defaulting to Autoproj.workspace) which is then used to resolve environment and configuration within the function.

This will also make the tests cleaner since they won't have to access the global stuff.

rock/python.rb
```diff
@@ -0,0 +1,66 @@
...
+                               "failed to find python")
+    end
+
+    # Allow to update the PYTHONPATH for package, tries to guess the python
+    # binary from Autobuild.programs['python'] and system's default setting
+    # @param[Autobuild::Package] pkg
+    # @param[String] python_bin Path to a custom python version
+    def self.activate_python(pkg, python_bin = nil)
+        begin
+            python_bin, python_version = Rock.find_python_bin(python_bin)
```
doudou: If I am passing a python bin, I assume that you will use it. You may error if some sanity check fails, but you should *definitely not* override it with something else.

A.k.a. split the "finding the bin", "finding the version" and "checking whether things are rosy" in find_python_bin and handle the multiple code paths either here in activate_python or in a separate `resolve_python` method. This will make the different cases (and how you react to them) a lot clearer. Right now, it's really complex to follow the control flow.

Additionally, one thing you might want to consider is that the python might be installed by Autoproj itself, which means that the existence check would fail on first bootstrap. Not sure what's best (assuming that python is installed prior to using Autoproj seems to be a reasonable workflow)

~~~
def self.resolve_python(ws: Autoproj.workspace, bin: ws.config.get('python_bin', nil), version: ws.config.get('python_version', nil))
    bin ||= find_python_bin
    if !bin
       # Error: can't find
    elsif !File.exist?(bin)
       # Error #{bin} does not exist
    end

    version ||= find_python_version(version)
    unless version
       raise "Error: can't detect version of #{bin}"
    end

    [bin, version]
end

def self.activate_python(pkg, ws: Autoproj.workspace, bin: ws.config.get('python_bin', nil), version: ws.config.get('python_version', nil))
     bin, version = resolve_python(ws: ws, bin: bin, version: version)
     ws.config.set('python_bin', bin)
     ws.config.set('python_version, version)
     ws.env.add_path ...
end
~~~

2maz:
I would also set install/bin/python in Rock.activate_python according to python_executable unless you think that would conflict with autoproj. Generally I would aim to move this functionality into autoproj anyhow.

2maz:
@doudou updated, could you recheck?

doudou:
> I would also set install/bin/python in Rock.activate_python according to python_executable unless you think that would conflict with autoproj. 

I don't think it will, no.

> Generally I would aim to move this functionality into autoproj anyhow.

I think we could move `activate_python` into autoproj, but I don't think it should auto-install it. The reason autoproj does it with Ruby is that it depends on Ruby itself (and definitely doesn't want people to use a different ruby that the one Autoproj has been installed with).

Having it in autoproj would allow people to call it in their `init.rb`

If you are going to do more python integration within Autoproj, it would make sense to have it integrated with a `virtualenv`-like package manager (the way BundlerManager does for Ruby). However, the landscape is a lot more complicated in Python-land ...

rock/python.rb
```diff
@@ -171,6 +171,22 @@ def self.resolve_python(ws: Autoproj.workspace,
         end
     end
 
+    def self.rewrite_python_shims(python_executable, root_dir)
+        shim_path = File.join(root_dir, "install","bin")
+        if !File.exist?(shim_path)
+            FileUtils.mkdir_p shim_path
+            Autoproj.warn "Rock.rewrite_python_shims: creating "\
```
doudou: Won't that hit at least one time for every workspace ?

You could avoid having that warning by issuing it only if `python_executable` was already set.

2maz: If I see it correctly in Autoproj::Workspace.setup at the stage of calling Rock.activate_python (now assuming init.rb) install/bin should be present. 
Hence, the warning will only show up if working with a corrupted or test workspace or outside an autoproj workspace.

doudou: Oh yes you're right. Misread it as if you were testing the whole file path, not just the directory. Sorry.

doudou: Anyways, all's good from my p.o.v.

2maz:
@doudou Thanks for your help, I opened a separate PR to discuss the activation part #156

PR #156, opened 2020-02-05
Add setup_python_configuration_options

2maz:
and allow to activate python

### DIFF
--- a/rock/python.rb
+++ b/rock/python.rb
@@ -170,6 +170,12 @@ module Rock
             return auto_resolve_python(ws: ws, version: version)
         end
     end
+    def self.remove_python_shims(root_dir)
+        shim_path = File.join(root_dir, "install","bin","python")
+        if File.exists?(shim_path)
+            FileUtils.rm shim_path
+        end
+    end
 
     def self.rewrite_python_shims(python_executable, root_dir)
         shim_path = File.join(root_dir, "install","bin")
@@ -200,6 +206,12 @@ module Rock
         [bin, version]
     end
 
+    def self.deactivate_python(ws: Autoproj.workspace)
+        remove_python_shims(ws.root_dir)
+        ws.config.reset('python_executable')
+        ws.config.reset('python_version')
+    end
+
     # Allow to update the PYTHONPATH for package, tries to guess the python
     # binary from Autobuild.programs['python'] and system's default setting
     # @param [Autobuild::Package] pkg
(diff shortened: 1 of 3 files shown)


rock/python.rb
```diff
@@ -200,6 +206,12 @@ def self.activate_python(ws: Autoproj.workspace,
         [bin, version]
     end
 
+    def self.deactivate_python(ws: Autoproj.workspace)
+        remove_python_shims(ws.root_dir)
+        ws.config.reset('python_executable')
+        ws.config.reset('python_version')
```
doudou: Should maybe save after removing the options.

rock/python.rb
```diff
@@ -216,4 +228,26 @@ def self.activate_python_path(pkg,
                              "python#{version}","site-packages")
         [bin, version, path]
     end
+
+    def self.setup_python_configuration_options(ws: Autoproj.workspace)
+        ws.config.declare 'USE_PYTHON', 'boolean',
+            :default => 'no',
```
doudou: Use "new" (2.0+) keyword arguments (default: 'no')

rock/python.rb
```diff
@@ -216,4 +228,26 @@ def self.activate_python_path(pkg,
...
+            :doc => [ "Do you want to activate python?" ]
+
+        value = ws.config.get('USE_PYTHON')
+        if value && !Autoproj::BuildOption::FALSE_STRINGS.include?(value)
```
doudou: 'boolean' should take care of the conversion. Doesn't it ?

2maz: It does not for autoproj <= v.2.11.0 when using noninteractive mode. Fortunately the bug has already been fixed for upcoming releases https://github.com/rock-core/autoproj/commit/efcd09dcd918281a3041b39a8f6defa3e2d7f2ea

rock/python.rb
2maz: Sorry, I do not understand what you are trying to say here,

doudou: Sorry, I thought I replied.

I am wondering if it wouldn't make sense to save the configuration file right away (`ws.config.save`). Just a wonder, really.

rock/python.rb
doudou: K. I need to release that stuff.

rock/python.rb
2maz: That would make an exemption from the rule - in all other cases config.save is triggered after all packages sets have been loaded. So I would not do it.

PR #157, opened 2020-02-18
Change cxx11 default to 'yes'

2maz:


doudou:
I don't like this as it affects all Rock packages, even those that explicitly select their C++ standard.

This flag never made any sense, and it still doesn't. Packages that want to use C++11 explicitly should have their CMake code say so. At the time where we created the flag, CMake had poor support for it. It now doesn't.

Just use [the CMake flag](https://www.rock-robotics.org/rock-and-syskit/libraries/cpp_libraries.html#cxx_standard)

doudou:
My vote: getting rid for the flag altogether. 18.04 and Debian 10 are both building with C++11 by default anyways.

2maz:
superseded by  #159

PR #159, opened 2020-02-18
Remove cxx11 option and leave setting of cxx standard

2maz:
to the individual configuration of a CMake package


doudou:
While we're at it ... should remove castxml support as well ? It's been a long time since gccxml has been available.

2maz:
You probably mean it the other way round (gccxml has been succeeded by castxml).
Is your intend to remove gccxml support from typelib then as well?

doudou:
> Is your intend to remove gccxml support from typelib then as well?

Unless you see a problem with that.

If you think it's still useful to keep gccxml support, we could also make castxml the default (as in: actually set the value to castxml if it is unset, so that the only way to change it would be to provide a seed config or set the value in one of the package set files)

2maz:
Since we still have some legacy systems in use, I would prefer to just change the default as you suggest to avoid extra trouble.

doudou:
> Since we still have some legacy systems in use, I would prefer to just change the default as you suggest to avoid extra trouble.

:+1:

The current pull request removes the flag, but doesn't set it to anything. Need to do something [like this](https://github.com/rock-core/package_set/pull/160/files#diff-242d0b4ac43f99fbb7c9023c1aca7166R163)

2maz:
Do we need that? I only renamed cxx11 to cxx  and setup_cxx_support is stilled called and setting typelib_cxx_loader?

doudou:
You removed the definition of the option, but didn't remove its usage. It then fails. It probably works because you have an already-bootstrapped install which still has the option in the config file.

For castxml, should we keep it as-is, or force-set by default to castxml (not asking the question anymore) ?

2maz:
We do not ask a question on castxml, but autodetect the right setup - so I would leave it as is.

You are right that I missed to remove the usage of the 'cxx11' flag in the overrides.rb - which should be fixed now.

doudou:
> We do not ask a question on castxml, but autodetect the right setup - so I would leave it as is.

Oh alright. Forgot that.

PR #161, opened 2020-03-11
Activate python extension only when USE_PYTHON is true

2maz:


rock/python.rb
```diff
@@ -212,30 +212,36 @@ def self.deactivate_python(ws: Autoproj.workspace)
...
                              bin: nil,
                              version: nil)
-        bin, version = resolve_python(ws: ws, bin: bin, version: version)
-        path = ws.env.add_path 'PYTHONPATH',
-                   File.join(pkg.prefix, "lib",
-                             "python#{version}","site-packages")
-        [bin, version, path]
+        if ws.config.has_value_for?('USE_PYTHON') and ws.config.get('USE_PYTHON')
+            bin, version = resolve_python(ws: ws, bin: bin, version: version)
+            path = ws.env.add_path 'PYTHONPATH',
```
doudou: I missed that one ... this should be `pkg.env_add_path`, not `ws.env.add_path`. The former ensures that the path gets added only if the package is actually part of the build.

rock/python.rb
```diff
@@ -212,30 +212,36 @@ def self.deactivate_python(ws: Autoproj.workspace)
...
     def self.activate_python_path(pkg,
                              ws: Autoproj.workspace,
                              bin: nil,
                              version: nil)
-        bin, version = resolve_python(ws: ws, bin: bin, version: version)
-        path = ws.env.add_path 'PYTHONPATH',
-                   File.join(pkg.prefix, "lib",
-                             "python#{version}","site-packages")
-        [bin, version, path]
+        if ws.config.has_value_for?('USE_PYTHON') and ws.config.get('USE_PYTHON')
```
doudou: Please stick to `&&` and `||` for booleans. `and` and `or` are seldom used in Ruby, and have different precedence than `&&` / `||`, which make them tricky.

doudou: I think this could simply be expressed as `ws.config.get('USE_PYTHON', false)`

rock/python.rb
```diff
@@ -212,30 +212,36 @@ def self.deactivate_python(ws: Autoproj.workspace)
...
+            bin, version = resolve_python(ws: ws, bin: bin, version: version)
+            path = ws.env.add_path 'PYTHONPATH',
+                       File.join(pkg.prefix, "lib",
+                                 "python#{version}","site-packages")
+            return [bin, version, path]
+        end
+        [nil,nil,nil]
```
doudou: Why not just `nil` ?

Also, would be easier on the reader to return nil first and then go through the rest of the code

~~~
return unless ws.config.get('USE_PYTHON', false)

# Do the rest
~~~

2maz:
Thx for the comments. Updated according to your suggestion.

PR #162, opened 2020-03-26
Use python venv for python setup

2maz:
Draft of a python venv integration:
@doudou do you have a suggestion on how to ensure availability of python3 osdeps, e.g., using ws.os_package_installer in rock-core's init.rb would require the osdeps files to be loaded already?

### DIFF
--- a/rock.osdeps
+++ b/rock.osdeps
@@ -288,6 +288,10 @@ python3-msgpack:
     debian,ubuntu: python3-msgpack
     opensuse: python3-msgpack
 
+python3-venv:
+    debian,ubuntu: python3-venv
+    opensuse: python3-virtualenv
+
 cython3:
     debian, ubuntu: cython3
     opensuse: python3-Cython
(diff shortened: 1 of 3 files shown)


doudou:
If you want to go down this route, I would very strongly suggest you integrate `venv` as a package manager in Autoproj directly, the way bundler manages Ruby gems. This would guarantee that all Python-related stuff is installed the way they should, and see one python workspace for all Python packages.

doudou:
If you want to go down this route, I would very strongly suggest you integrate `venv` as a package manager in Autoproj directly, the way bundler manages Ruby gems. This would guarantee that all Python-related stuff is installed the way they should, and see one python workspace for all Python packages.

doudou:
> Draft of a python venv integration:
> @doudou do you have a suggestion on how to ensure availability of python3 osdeps, e.g., using ws.os_package_installer in rock-core's init.rb would require the osdeps files to be loaded already?

I don't think there is an easy way. At this stage, I really believe it would be best to convert this integration into a proper Autoproj package manager / package type, which would allow to (1) create and configure the python workspace for the whole autoproj workspace and (2) do the package-related stuff in different stages.

2maz:
Ok, then I will start looking into the autoproj integration.

PR #169, opened 2020-05-11
frame_helper: automatically set OPENCV4 if the `opencv4` pkg-config file can be found

doudou:
Depends on:
- [ ] https://github.com/rock-core/perception-frame_helper/pull/13


doudou:
This is superseded by the automatic detection at CMake level (https://github.com/rock-core/perception-frame_helper/pull/13)

PR #170, opened 2020-05-12
Qt4 support in Ubuntu 20.04

doudou:
Qt4 is dead, we knew that for a long time and I procrastinated working
on a solution alternative to the QtRuby bindings (did some work this
year, but too limited). Now, there's a wall because people stop
supporting it (for good reason)

This PR works around the problem. We'll have to have a solution for 22.04.


doudou:
Gents, I'm putting this here so that everyone knows it exists. This is very much WIP.

g-arjones:
Glad to know you are already looking into this. Also, it's very brave of you to commit to maintaining an apt repository :P

doudou:
> Also, it's very brave of you to commit to maintaining an apt repository :P

Well ... that's the good part of Ubuntu LTS. Once it's built for it, you shouldn't have to change it (and PPAs exist).

doudou:
Seems to work reasonably well. I've tested the Syskit IDE (so, basically vizkit) and tried some 3D visualization for the vizkit3d part.

doudou:
Ping @g-arjones @2maz

Gemfile
```diff
@@ -0,0 +1,2 @@
+gem "qtbindings", git: "https://github.com/rock-core/qtbindings"
```
g-arjones, line 1: Doesn't this mean qtbindings will be installed whether it is required or not?

doudou: Oh my ... you're absolutely right. And it uses the git repo for everything *included* on 18.04 where it's not needed.

The only way I see to handle this would be to add gem-as-git support within the osdeps subsystem. Would you see any other ?

g-arjones: Nope... I'm just not sure whether I would add support for that in osdeps files or if would rather use the Bundler management thing you implemented recently

doudou: > would rather use the Bundler management thing you implemented recently

Not sure what you are referring to.

g-arjones: I meant you could have a method like [this one](https://github.com/rock-core/autoproj/blob/25b1f1a31618fd75689a35ebcdf1dcbd7b72a250/lib/autoproj/package_managers/bundler_manager.rb#L163) to set repositories, branches

doudou: Mmmm. Interesting. We do not get the per-OS handling, though (to avoid overriding the gem when not on Focal)

g-arjones: You do if you implement it to do so :-P
I mean, one can do it in the script files using methods from the [OSPackageResolver](https://github.com/rock-core/autoproj/blob/master/lib/autoproj/os_package_resolver.rb) to get the guessed OS or you could provide an API in bundler manager for that...

doudou: Yeah. In that case I think it will be cleaner to just add support for it from the osdeps...

g-arjones: The API in `BundlerManager`, could store in the configuration something like an `only_in` setting that it would check against the guessed OS before passing it along to bundler

g-arjones: Nothing against doing it in osdeps file though, I just don't know how it is going to look

wieset:
Hi! I wanted to check about the status of this pull request and of Ubuntu 20.04 compatibility in general. Any chance to have Rock ready for 20.04 soon? I don't know what your plans are internally, but on our side we would like to upgrade our systems in the near future, so that would be great!

doudou:
Hi @wiesel. The problem with this pull request is backward compatibility. We need to implement something in Autoproj to handle the gem-as-git qtbindings gem before we can merge.

A.k.a. with this pull request in, 20.04 works fine. I've been using it for a while.

This is not ideal, but I myself will probably not have time to work on it for a little while still - like a month or so

levingerdes:
Hi @doudou, @wieset and I found two more problems while trying this solution:
- there is a dependency on `ruby2_keywords` which we could only find as `ruby-ruby2-keywords` in the Ubuntu 20.04 repositories. To continue testing I've added the `ruby/ruby2_keywords` github package manually.
- `libqwt5-qt4-dev` had to be set to `nonexistent` for Ubuntu 20.04 in both `rock-core/package_set` and `rock-core/rock-package_set`: https://github.com/rock-core/rock-package_set/pull/231/commits/62033f54966a426cb0bf8d61877369c3203f292c

doudou:
> there is a dependency on ruby2_keywords which we could only find as ruby-ruby2-keywords in the Ubuntu 20.04 repositories. To continue testing I've added the ruby/ruby2_keywords github package manually.

Ah good. Indeed. It's available (and should be installed) as a gem.

> libqwt5-qt4-dev had to be set to nonexistent for Ubuntu 20.04 in both rock-core/package_set and rock-core/rock-package_set: rock-core/rock-package_set#231

Oh. Thanks for this one. I am using gazebo, and had to get rid of this package since forever because of that (conflicts).

doudou:
@levingerdes actually, ruby2_keywords is already defined as a gem [in rock.core](https://github.com/rock-core/package_set/blob/master/rock.osdeps). Maybe you had to look for it before it got defined ?

wieset:
@doudou The reason for this problem was that the `qt4_focal` branch is behind the respective commit https://github.com/rock-core/package_set/commit/cb2b2d78e6e98518d9cfb7fc0684e742a9d1dcbb. But I saw that you created a new `focal` branch rebased on master, which we are now using.

wieset:
@doudou We've created some pull requests for issues we found while trying to use this branch. Would be great if you could have a look (and potentially merge & rebase again).

https://github.com/rock-core/rock-package_set/pull/232
https://github.com/rock-core/package_set/pull/181
https://github.com/rock-core/package_set/pull/182

PR #171, opened 2020-05-13
explicitly mark python-msgpack as nonexistent on 20.04

doudou:
The package that depends on it (pocolog2msgpack) will have to
explicitly upgrade to the Python3 version. At least, until then,
Rock Core bootstraps fine (it is an optional dependency of
pocolog2msgpack)

### DIFF
--- a/rock.osdeps
+++ b/rock.osdeps
@@ -265,7 +265,9 @@ python-pexpect:
     opensuse: python-pexpect
 
 python-msgpack:
-    debian,ubuntu: python-msgpack
+    debian,ubuntu:
+        "16.04,18.04,18.10,19.04,19.10": python-msgpack
+        default: nonexistent
     opensuse: python-msgpack
 
 python3:


doudou:
See also https://github.com/rock-core/tools-pocolog2msgpack/issues/5

rock.osdeps
```diff
@@ -265,7 +265,9 @@ python-pexpect:
     opensuse: python-pexpect
 
 python-msgpack:
-    debian,ubuntu: python-msgpack
+    debian,ubuntu:
+        "18.04,18.10,19.04,19.10": python-msgpack
```
g-arjones, line 269: Could you please add 16.04 here as well?

rock.osdeps
```diff
@@ -265,7 +265,9 @@ python-pexpect:
...
+        default: nonexistent
```
g-arjones, line 270: Maybe install from pip instead?

doudou: It's not that the package does not exist, it is that it's only available for Python 3.

I'd like to leave the Python 2 / Python 3 migration strategy to the people that actually are dealing with Python, @2maz ?

g-arjones: > it is that it's only available for Python 3

That's my point... Can't you install the python2 version from pip?

doudou: I don't want to build installing a python2 package from pip within the rock-core package set, unless there's something stepping in (i.e. not me) that vouches and maintains it.

I know what mixing versions of stuff does - in both compiled and dynamic languages - and I don't want to be responsible for it.

rock.osdeps
doudou: Done

rock.osdeps
g-arjones: > I don't want to be responsible for it

Fair enough

rock.osdeps
g-arjones: Thanks

PR #173, opened 2020-05-25
Add fallback shell detection to 04stable.autobuild for base/scripts

0Nel:
In some use cases (especially in docker setups) the environment SHELL variable might not be set. This PR adds a fallback option, providing a stable alternative to acquire the shell type via a system call.

### DIFF
--- a/04stable.autobuild
+++ b/04stable.autobuild
@@ -98,11 +98,7 @@ in_flavor 'master', 'stable' do
                 else Autoproj.shell_helpers?
                 end
 
-            unless shell = ENV['SHELL']
-                Autoproj.warn("SHELL Variable not set, using fallback detection")
-                shell ||= %x[ps -p $$ -ocomm= | xargs which].strip
-            end
-            if shell_helpers
+            if shell_helpers && shell=ENV['SHELL']
                 shell_kind = File.basename(shell)
                 if shell_kind =~ /^\w+$/
                     shell_file = File.join(pkg.srcdir, "shell", shell_kind)
@@ -119,6 +115,8 @@ in_flavor 'master', 'stable' do
                         end
                     end
                 end
+            elsif shell.nil?
+                Autoproj.error("Unable to detect shell type. Please set the SHELL environment variable.")
             end
         end
     end
--- a/04stable.autobuild
+++ b/04stable.autobuild
@@ -99,6 +99,9 @@ in_flavor 'master', 'stable' do
                 end
 
             if shell_helpers && shell = ENV['SHELL']
+                if shell.nil?
+                    Autoproj.warn("SHELL Variable not set, using fallback detection")
+                    shell ||= %x[ps -p $$ -ocomm= | tr -d '\n']
                 shell_kind = File.basename(shell)
                 if shell_kind =~ /^\w+$/
                     shell_file = File.join(pkg.srcdir, "shell", shell_kind)
--- a/04stable.autobuild
+++ b/04stable.autobuild
@@ -98,10 +98,11 @@ in_flavor 'master', 'stable' do
                 else Autoproj.shell_helpers?
                 end
 
-            if shell_helpers && shell = ENV['SHELL']
-                if shell.nil?
-                    Autoproj.warn("SHELL Variable not set, using fallback detection")
-                    shell ||= %x[ps -p $$ -ocomm= | tr -d '\n']
+            unless shell = ENV['SHELL']
+                Autoproj.warn("SHELL Variable not set, using fallback detection")
+                shell ||= %x[ps -p $$ -ocomm= | tr -d '\n']
+            end
+            if shell_helpers
                 shell_kind = File.basename(shell)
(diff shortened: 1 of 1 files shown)


04stable.autobuild
```diff
@@ -98,7 +98,11 @@ in_flavor 'master', 'stable' do
                 else Autoproj.shell_helpers?
                 end
 
-            if shell_helpers && shell = ENV['SHELL']
+            unless shell = ENV['SHELL']
+                Autoproj.warn("SHELL Variable not set, using fallback detection")
+                shell ||= %x[ps -p $$ -ocomm= | tr -d '\n']
```
2maz, line 103: ENV['SHELL'] holds the full path, while `shell` now contains only the basename. So I would suggest to use `%x[ps -p $$ -ocomm= | xargs which].strip`

0Nel: I thought this doesn't pose a problem as the immediate next action on the variable is stripping it to its base name (see line 106). However I do agree that your proposed edit is cleaner. Going to edit this accordingly..

doudou:
What about raising and telling the user to set SHELL before calling autoproj ?

This feels like really trying too hard.

0Nel:
I understand that there are other ways to approach and solve this issue. However, what I do not understand is, why it is preferable to bother the user, if there is a clean and stable solution at hand?

doudou:
> if there is a clean and stable solution at hand?

The issue here is that we don't have the same definition of "clean and stable solution". This in my book is a hacky heuristic.

Environments where SHELL is not defined are not environments I'm expecting any sort of user to encounter. So, yes, I prefer to let the advanced developer set its SHELL variable and/or **explicitly** tell autoproj what to do (through e.g. a configuration variable) rather than doing ... this.

What would be a welcome addition is a clean and stable error message that tells the user what is the problem and what he/she should do about it.

0Nel:
Ok, I have to admit that I am not coming from a computer science background. Thus I acknowledge that you have a better idea of what is clean and stable. Just to improve my book; is it hacky cause it uses shell commands go get the shell type? Is it a heuristic cause it is not necessarily stable across a broad number of shells and OS?

I will adapt the PR accordingly and have it print out a clean and stable error message ;) I actually think that due to the really concise docker setup a number of developers might encounter this problem. However a clear error message will probably be help enough..

04stable.autobuild
```diff
@@ -115,6 +115,8 @@ in_flavor 'master', 'stable' do
                         end
                     end
                 end
+            elsif shell.nil?
+                Autoproj.error("Unable to detect shell type. Please set the SHELL environment variable.")
```
doudou, line 119: " or disable shell helpers altogether by setting the `shell_helpers` configuration variable to false"

doudou:
Thanks for the understanding, @0Nel

It is a heuristic because it is trying to guess information from a side channel (i.e. not getting it from a primary source of information). In this case, you're trying to guess which shell the user uses by running a command and hoping it will be run in the same shell. Nothing guarantees that last part (it actually is guaranteed to not be, see below). Just went to see [Wikipedia's definition of heuristic](https://en.wikipedia.org/wiki/Heuristic), and I like it quite a lot

(also, I find the concept of shelling out rather bad-looking in general. Now, we're trying to handle a shell, so in this particular case, I don't think it would be too bad)

All heuristics have failure modes. So, as with all heuristics, you have to balance the number of people impacted vs. the possibility that it will fail - because once the heuristic is there, you have to maintain it.

That's for the general principles.

Now, some failure modes:
- you haven't considered which shell Ruby uses to run the command itself (Ruby has to pick one, and ruby is software a.k.a. not magical mind-reading). It seems that Ruby always uses `sh` for the backticks (from [stackoverflow](https://stackoverflow.com/questions/12629519/which-shell-ruby-uses-as-its-subshell), and verified on my shell, but I haven't been looking at the source code). So, you heuristic is essentially returning `/bin/sh` on Linux.
- a slightly better heuristic would have been to look for the shell that called autoproj. But, it's harder and has its own failure modes (like, not actually having a shell calling autoproj)
- a bit more minor, busybox's `ps` does not have `-p`. Given that the primary target for this patch are containers, I'd argue that being compatible with busybox would not be a horrible idea

0Nel:
Thanks for your detailed answer! Really gives me an opportunity to rethink and improve my coding style! Also very interesting hint that ruby by default is using the sh shell for system calls!

Considering the possible or even inevitable failure modes I agree that printing an error message is the way to go.

04stable.autobuild
```diff
@@ -98,7 +98,7 @@ in_flavor 'master', 'stable' do
                 else Autoproj.shell_helpers?
                 end
 
-            if shell_helpers && shell = ENV['SHELL']
+            if shell_helpers && shell=ENV['SHELL']
```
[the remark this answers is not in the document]
0Nel: At least some minor code improvement with this PR ;)

doudou:
Thanks @0Nel !

PR #177, opened 2020-07-10
define the ruby2_keywords gem as an osdeps

doudou:
This defines ruby2_keywords for ruby < 2.7, which helps fixing an
incompatibility moving forward with the keyword arguments.

As Gabriel said, the Ruby devs really messed this up

### DIFF
--- a/rock.osdeps
+++ b/rock.osdeps
@@ -335,3 +335,4 @@ rubocop:
     gem:
         - rubocop=0.83.0
 rubocop-rock: gem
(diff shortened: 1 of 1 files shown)


doudou:
Forgot that https://github.com/rock-core/tools-syskit/pull/254 was depending on this, and merged it.

I think this is fairly straightforward. I'm going to merge it too, I'll revert both if you find some issues.

Apologies.

PR #181, opened 2020-10-30
Unify qwt5 osdep from rock-package_set

wieset:
`WARN: osdeps definition for qwt5, previously defined in /home/user/rock/autoproj/remotes/rock.core/rock.osdeps overridden by /home/user/rock/autoproj/remotes/rock/rock.osdeps`

The qwt5 osdep is defined in mutiple package sets. I would suggest to consolidate it here and remove the definition from the rock package set.

### DIFF
--- a/rock.osdeps
+++ b/rock.osdeps
@@ -75,8 +75,11 @@ rice:
         - ruby-dev
 
 qwt5:
-    debian,ubuntu: libqwt5-qt4-dev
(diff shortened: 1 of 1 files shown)


wieset:
See also https://github.com/rock-core/rock-package_set/pull/232

doudou:
Thanks @wieset

PR #183, opened 2021-01-11
Organize python osdeps in version specific files

2maz:
To facilitate switching from 2 to 3 existing python-X definitions will point to corresponding python3 packages.

### DIFF
--- a/rock.osdeps
+++ b/rock.osdeps
@@ -255,56 +255,6 @@ ruby-backports:
 ruby-xmlrpc:
    gem: xmlrpc
 
-python:
-    debian,ubuntu: python-dev
-    opensuse: python-devel
-
-python-nose:
-    debian,ubuntu: python-nose
-    opensuse: python-nose
-
-python-pexpect:
-    debian,ubuntu: python-pexpect
-    opensuse: python-pexpect
-
-python-msgpack:
-    debian,ubuntu:
-        "16.04,18.04,18.10,19.04,19.10": python-msgpack
-        default: nonexistent
-    opensuse: python-msgpack
-
-python3:
-    debian,ubuntu: python3-dev
-    opensuse: python3-devel
-
-python3-pip:
-    debian,ubuntu: python3-pip
-    opensuse: python3-pip
-
-python3-nose:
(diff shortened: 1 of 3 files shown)


doudou (approved):

g-arjones (approved):

wieset:
This has broken our build. Python osdeps in the `rock.osdeps-pythonX` files are not found by autoproj, whereas they were found in the `rock.osdeps` file. Can you confirm that using files that don't strictly end in `.osdeps` is supposed to work? Any idea why it might not?

`osdeps reports: cannot resolve python3: python3 is not an osdep and it cannot be resolved as a source package`

doudou:
@wieset I think the mistake was in https://github.com/rock-core/package_set/pull/180. It essentially assumes that all people that use Python would be calling `activate_python`. Which rather obviously in hindsight could not be the case ...

We could move python2/python3 to the main osdeps since they are versioned. But maybe you are depending on other dependencies as well ?

I think the best would be to call activate_python in your package set's `init.rb`

wieset:
Thanks @doudou. In this case, the system python version is 2.7 but we have one package that needs `python3-dev` installed. So calling `Rock.activate_python` doesn't do the trick since it only adds the python2 osdeps file.

We could add `python3` to our own osdeps file, but then we risk a double definition in the future. I guess we could use a different name for the osdep to avoid that. Or you move python3 to the main osdeps as you say.

2maz:
@wieset I would suggest that you add the python3-dev that you need in your default python2 setup into a x.osdeps-python2 of your current package set.

wieset:
@2maz Ok sounds good, we'll do that!

2maz:
The current setup of python in autoproj is performing a clean separation of python2 and python3 osdeps, but you are having a mixed setup. Apparently I wrongly assumed that installations will run on either python2 or python3.

Regarding the setup: no extra call to activate_python should be needed.
Including rock.core package_set will add configuration options and ask for the activation via: https://github.com/rock-core/package_set/blob/3bb28297f3003eeee2e88f5a3ae6b923878f408a/init.rb#L152
should

PR #184, opened 2021-01-13
Update python version for internal use

2maz:
Handle the case, when autoproj is started in the context of a different python version than
the selected one.

### DIFF
--- a/rock/python.rb
+++ b/rock/python.rb
@@ -204,11 +204,13 @@ module Rock
                 "are you operating on a valid autoproj workspace?"
         end
 
-        File.open(File.join(shim_path, 'python'), 'w') do |io|
+        python_path = File.join(shim_path, 'python')
+        File.open(python_path, 'w') do |io|
             io.puts "#! /bin/sh"
             io.puts "exec #{python_executable} \"$@\""
         end
-        FileUtils.chmod 0755, File.join(shim_path, 'python')
+        FileUtils.chmod 0755, python_path
+        python_path
     end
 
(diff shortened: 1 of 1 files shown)


doudou:
> Handle the case, when autoproj is started in the context of a different python version than
the selected one.

Autoproj itself does not set these AFAIK. If you found out that they changed, I'm guessing it's because other parts of some package sets have set them.

Generally speaking, being able to have some parts of the install use one and some other parts another will probably lead to .... interesting situations.

2maz:
You might be right, it's probably a different issue I have to fix -- I will recheck.

2maz:
It turned out to be a misconfiguration of our internal CI - where some base image was using a seed config with python_executable set to /usr/bin/python, which is python2.7 by default on Ubuntu 18.04.
Thanks for your response and sorry for the noise.

doudou:
Well. Maybe you'd want to add a sanity check about this ?

2maz:
Not really sure what a sanity check would mean here?
Everything works basically as expected - if you want /usr/bin/python and do not request a particular version then inconsistencies are exposed as soon as you have a package that depends only on python3.

However, @planthaber suggested to set the default of python to /usr/bin/python3, which I would agree would make sense since among other reasons this is now the only supported python (major) version: see #185 for the suggested change

doudou:
> Not really sure what a sanity check would mean here?

Making sure that a version being selected and then another one would raise. It seemed to me that it was the issue you had here. Now, if it's not just disregard my remark :P

PR #186, opened 2021-01-19
ruby2_keywords 0.0.3 is for ruby 2.6+ only, stick to 0.0.2

doudou:


doudou:
@2maz @g-arjones I can confirm it does what it should in CI on Ubuntu 18.04 and Ubuntu 20.04 does not have the problem. This breaks anything that loads Syskit (not uses, only load). I'm merging now.

PR #189, opened 2021-02-08
create tools/orogen-syskit_tests

doudou:
This is a test-only package for the benefit of Syskit's test suite.

Syskit currently has no test that actually starts orogen-generated
components. This will provide it with what is necessary to write
some.

### DIFF
--- a/00new_packages.autobuild
+++ b/00new_packages.autobuild
@@ -6,6 +6,9 @@ remove_from_default 'tools/apaka-rock_patches'
 ruby_package 'tools/apaka'
 remove_from_default 'tools/apaka'
 
+orogen_package "tools/orogen_syskit_tests"
+remove_from_default "tools/orogen_syskit_tests"
+
 cmake_package 'tools/msgpack-c' do |pkg|
     pkg.define "MSGPACK_BUILD_TESTS", pkg.test_utility.enabled?
 end


00new_packages.autobuild
```diff
@@ -6,6 +6,9 @@ remove_from_default 'tools/apaka-rock_patches'
 ruby_package 'tools/apaka'
 remove_from_default 'tools/apaka'
 
+orogen_package "tools/orogen-syskit_tests"
```
g-arjones, line 9: Doesn't this name violate rock's package naming conventions? Folders separated by hyphens and words separated by underscores...

It looks weird.

doudou: I'm open to suggestions, but tools/orogen/syskit_tests is not possible since `tools/orogen` is a package.

g-arjones: `tools/orogen_syskit_tests` or `tools/syskit_orogen_tests` ? The name itself doesn't make much of a difference to me really, the one thing that looks out of place is the hyphen because of the naming convention.

That said, I'm not even sure there is really a convention or if that's just a habit so you might just as well leave it as-is if there's not.

doudou: I think I unconsciously picked `orogen-syskit_tests` because then the repositories match the name they should have (`tools-orogen-syskit_tests`). But I agree that it's actually confusing.

`tools/syskit_orogen_tests` sounds good.

g-arjones (approved):

00new_packages.autobuild
doudou: Just for the record, I picked orogen_syskit_tests in the end as syskit_orogen_tests sounded too much like "tests for orogen that are part of syskit".

PR #195, opened 2021-10-05
`prefix_dir` isn't always `root_dir/install`

moooeeeep:
When we would specify a non-default install prefix with Python activated, the Python binaries would still be placed at `install/bin`. While everything else gets installed to the actual `prefix_dir`. This is supposed to address this. 

I have a similar PR for autoproj, to which I'll add a link. Please review and merge if appropriate.

### DIFF
--- a/rock/python.rb
+++ b/rock/python.rb
@@ -189,15 +189,15 @@ module Rock
             return auto_resolve_python(ws: ws, version: version)
         end
     end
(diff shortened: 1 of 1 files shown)


doudou (approved):

doudou:
@moooeeeep is this something we should still merge ? Or the autoproj patch is the one ?

I'm rather lost w.r.t. python support here ...

PR #200, opened 2022-01-11
Set github default protocol

planthaber:
Github disabled the git protocol access, so http should be the default.


planthaber:
Seems like thsi wasn't the issue, which was only temporary, closing.

PR #202, opened 2022-02-11
Add qwt for focal to qwt5 osdeps

0Nel:
Indeed qwt5 for qt4 is `nonexistent` on ubuntu 20.04, which requires qt5.

However packages that require qwt on ubuntu 20.04 build fine with `libqwt-qt5-dev`

At first qwt5 did not seem to be the correct key, however since qwt is defined for fedora, opensuse and darwin as well I didn't open up a new entry.

### DIFF
--- a/rock.osdeps
+++ b/rock.osdeps
@@ -77,7 +77,7 @@ rice:
 qwt5:
     debian,ubuntu:
         default: libqwt5-qt4-dev
-        '20.04': nonexistent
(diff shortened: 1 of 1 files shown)


planthaber:
So there should be a new entry
```yaml
qwt6:
    debian,ubuntu:
        default: libqwt-qt5-dev
```

or we create an additional qwt:  dependency in case qwt5 and 6 are compatible. For pacages that qupport qt4 and 5

PR #203, opened 2022-02-14
hard-fork orocos-toolchain packages we need for rock

doudou:
I've been unable to reach anyone on the orocos side for 2-3 years now
to discuss the future of ... anything really. Even RTT. I think it's time
to stop bothering.

### DIFF
--- a/source.yml
+++ b/source.yml
@@ -94,14 +94,18 @@ overrides:
     - ^(orogen|typelib|rtt|utilrb|utilmm|rtt_typelib|tools/metaruby)$:
       branch: $ROCK_BRANCH
 
-    # pin rtt commit to before the merge of the last major dataflow refactoring
-    #
-    # This refactoring led to massive deadlocking on a lot of dynamic
-    # disconnection scenarios.
-    #
-    # See https://github.com/orocos-toolchain/rtt/pull/302
     - rtt:
       github: rock-core/rtt
+    - orogen:
+      github: rock-core/tools-orogen
+    - typelib:
+      github: rock-core/tools-typelib
+    - utilrb:
+      github: rock-core/tools-utilrb
+    - rtt_typelib:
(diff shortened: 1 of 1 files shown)


doudou:
Good questions

1. because there are still people using the orocos toolchain, who therefore happen to depend on the typelib version in orocos-toolchain/typelib. As I don't plan to support this use case (which was always not very well working in the first place), I don't want to work there. This makes it clear that I consider that typelib is a rock-core project (which, in hindsight, it always was) maintained by Rock developers.
2. because without removing the linkage, pull requests and issues end up in the original repository, which creates a lot of confusion.

g-arjones (approved):

doudou:
If there is an agreement of principle, I'll change this pull request to hard-fork all the packages that are relevant to Rock, that is, in addition to rtt (which is already forked) and typelib:

- utilrb
- rtt_typelib

@planthaber, thoughts ? Are you the right person to discuss this with at DFKI ? If not, can you point me to that person ?

planthaber:
> @planthaber, thoughts ? Are you the right person to discuss this with at DFKI ? If not, can you point me to that person ?

I'm fine with this, and yes, i am the right person now.

planthaber (approved):

planthaber:
What do you think of renaming the packages to match the folder names?

e.g. orogen->tools/orogen and add metapackage to match also the old name?

Thsi whould make stuff like 'amake tools/orogen' actually work instead of `amake orgoen`

We could remove the moving of the packages here:
https://github.com/rock-core/package_set/blob/master/01import_orocos_toolchain.autobuild#L51

doudou:
I want to go slow on this, but that's on my list, yes.

For me, the cleanup steps are:
- remove the flavor system (incl. the OCL question)
- remove the dependency on orocos.toolchain, renaming the packages to match our scheme. Add aliases from the old name to the new names.

PR #204, opened 2022-03-16
use http as default protocol

planthaber:
Github disabled git protocol, so select http,ssh by default.
Pushes using http now needs token creation, so I think most are using ssh anyways


planthaber:
I just realized that the issue were (old) seed configs on our buildserver, sorry.

PR #206, opened 2022-05-04
add qt5-svg required by envire-core

annaborn:
@planthaber Steffen, could you please check the commit and merge it? :)

### DIFF
--- a/rock.osdeps
+++ b/rock.osdeps
@@ -130,6 +130,10 @@ qt5:
     macos-port: qt4-mac
     arch,manjarolinux: qt4
 
+qt5-svg:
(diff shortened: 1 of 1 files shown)


planthaber:
If it in only envire_core, it should be in the envire package set:
https://github.com/envire/envire-package_set/blob/master/envire.osdeps

annaborn:
Thank you! I will check. If it is only for envire, I will move it to the corresponding package.

pierrewillenbrockdfki:
Did a quick grep, i know of these packages that use qt svg features: planning/templ, tools/graph_analysis, envire/envire_core. On a side node, q

PR #215, opened 2022-10-26
feat: add tools/runkit

doudou:
It is a cleaned up, minimal fork or orocos.rb that is designed to support Syskit. tools/orogen_runkit_tests is a supporting package for runkit's test suite.


jhonasiv (approved):

jhonasiv (approved):

jhonasiv (approved):

doudou:
As always ... major refactorings don't work. Will re-do that work in a much more incremental way.

PR #217, opened 2023-01-11
Feature/qt5mergable

pierrewillenbrockdfki:
This PR merges the osdeps changes from the feature/qt5 branch into the master branch and then creates a configuration variable "QTVER" to set the qt version to be used by the packages that potentially link together possibly through use of vizkit3d. Finally, the configuration variable is used to setup vizkit3d to compile against the given qt version.

### DIFF
--- a/00new_packages.autobuild
+++ b/00new_packages.autobuild
@@ -33,3 +33,6 @@ end
 
 cmake_package "gui/osg_qt4"
 remove_from_default "gui/osg_qt4"
+
+cmake_package "gui/osg_qt5"
+remove_from_default "gui/osg_qt5"
--- a/04stable.autobuild
+++ b/04stable.autobuild
@@ -211,6 +211,8 @@ in_flavor 'master', 'stable' do
         pkg.env_add_path 'QT_PLUGIN_PATH', File.join(pkg.prefix, "lib", "qt")
     end
 
+    cmake_package 'gui/qtpropertybrowser'
+
     ##### Base implementation for data acquisition
     cmake_package 'drivers/iodrivers_base'
     orogen_package 'drivers/orogen/iodrivers_base'
--- a/04stable.autobuild
+++ b/04stable.autobuild
@@ -6,10 +6,17 @@ manifest = if Autoproj.respond_to?(:workspace)
 add_packages_to_flavors 'stable' => ['orogen', 'rtt', 'utilmm', 'utilrb', 'typelib', 'rtt_typelib', 'tools/metaruby', 'ocl', 'log4cpp']
 
 in_flavor 'master', 'stable' do
-    ruby_package 'gui/vizkit'
+    if Autoproj.config.get('QTVER') == '4'
+        ruby_package 'gui/vizkit'
+    end
 
     cmake_package 'gui/vizkit3d' do |pkg|
-        pkg.depends_on "gui/osg_qt4"
+        # gui/vizkit3d needs to do this in their manifest
+        if Autoproj.config.get('QTVER') == '4'
+            pkg.depends_on "gui/osg_qt4"
+        else
+            pkg.depends_on "gui/osg_qt5"
+        end
         if manifest.package_enabled?('rtt', false) # the toolchain is built, add it to the rtt_target
             pkg.define "OROCOS_TARGET", Autoproj.config.get('rtt_target')
         end
--- a/04stable.autobuild
+++ b/04stable.autobuild
@@ -6,8 +6,6 @@ manifest = if Autoproj.respond_to?(:workspace)
 add_packages_to_flavors 'stable' => ['orogen', 'rtt', 'utilmm', 'utilrb', 'typelib', 'rtt_typelib', 'tools/metaruby', 'ocl', 'log4cpp']
 
 in_flavor 'master', 'stable' do
-    ruby_package 'gui/vizkit'
-
     cmake_package 'gui/vizkit3d' do |pkg|
         pkg.depends_on "gui/osg_qt4"
         if manifest.package_enabled?('rtt', false) # the toolchain is built, add it to the rtt_target
--- a/init.rb
+++ b/init.rb
@@ -170,3 +170,53 @@ unless Autoproj.config.has_value_for?('syskit_use_bundles')
     Autoproj.config.set 'syskit_use_bundles', true, true
 end
 
+# Determine the qt default version depending on the distribution version.
+# The default here is 5 as the catch-all, but we actually prefer QT4
+# where available. It just won't be available on anything newer than the
+# versions given here. distributions with rolling releases probably
+# have QT4 removed already.
+os_names, os_versions = Autoproj.workspace.operating_system
+
+qt4_max_versions = [["ubuntu", 18], ["debian", 10], ["fedora", 38], ["opensuse", 15]]
+
+qtdefault = '5'
+qt4_max_versions.each do |(osname,maxversion)|
+    if os_names.include?(osname)
+        os_versions.each do |ver|
+            ver.match /^(\d+)\./ do |m|
+                if m[0].to_i <= maxversion
+                    qtdefault = '4'
+                end
+            end
+            ver.match /^(\d+)$/ do |m|
+                if m[0].to_i <= maxversion
+                    qtdefault = '4'
+                end
+            end
+        end
+    end
+end
+
+# The PPA for Ubuntu 20.04 only has x86_64 versions
+only_on ["ubuntu","20.04"] do
+    if RUBY_PLATFORM.downcase.include?("x86_64")
+        qtdefault = '4'
+    end
+end
+
+# This is a string since the ROS derived condition= attribute of
+# packages manifest.xml <depend> tag only deals in strings. Conversion
+# from int would be automatic, but this way we don't even pretend to be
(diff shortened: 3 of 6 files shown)


pierrewillenbrockdfki:
Next on my agenda would be to merge vizkit3ds feature/qt5 back into its master, using the configuration variable to make it require the correct dependencies and provide it a cmake configuration so it can look for the correct qt version.

If the consensus is to keep this work on a branch(for example feature/qt5), that can easily be arranged, but i think it is easier to discuss/review this PR relative to master, and i think this basically is ready to be merged.

So bikeshed away!

@doudou @planthaber

doudou:
> So bikeshed away!

That's my cue I guess.

First of all, thanks for the hard work.

Unfortunately, this work does stress the two rock "forks" (or "flavors" if I want to be gentler). The syskit-using people and the orocos-cpp-using people. The first one need qt4. The second one probably does not "need" but should be "strongly encouraged to use" qt5. Presenting the choice as "which version of Qt do you want" is not the right question. What should be asked is "which orogen-based toolchain do you want to use" ? I don't think it's a question that should be asked.

How to solve this ? Good question. I've been bothered thinking with this situation for quite a bit of time. One way, which probably would be the cleanest, would be to actually split. Have a common package set with the packages, and create metapackages / config options and buildconfs to have people "choose". I'm open to comments and other suggestions.

pierrewillenbrockdfki:
> Unfortunately, this work does stress the two rock "forks" (or "flavors" if I want to be gentler). The syskit-using people and the orocos-cpp-using people. The first one need qt4. The second one probably does not "need" but should be "strongly encouraged to use" qt5. Presenting the choice as "which version of Qt do you want" is not the right question. What should be asked is "which orogen-based toolchain do you want to use" ? I don't think it's a question that should be asked.

In my opinion, this should not be about the toolchain, because if qtruby bindings for qt5 would appear, the qt version problem does not go away, but the toolchain issue does.

Maybe we can adjust the current wording of the qt choice to be clearer and also mention syskit?

Anyway, again IMO, syskit will not survive if it cannot find a way to port towards qt5(and later, qt6), or drop the qt dependency. Operating systems will not make it easy to keep qt4 compiling and running, and Ubuntu 20.04 will not be supported forever. 

For porting towards qt5, one would have to try to get (some replacement of) qtruby to work for qt5, and maintain that version. Unfortunately, that is currently beyond my skills(at least from what i have seen), otherwise i would have tried that. I also wondered if one could just provide a minimum set of qt operations and do the conversion manually, but people have built complex GUIs with qtruby(including rock-display), using big chunks of qts functionality, so one would really want the whole thing. Not saying that doing the bindings manually is not an option, that just sounds like a lot of work. Also, even this would need changes in the existing rock ruby codebase; code supporting both would have to be aware that it needs to do things differently in some few places, depending on the qt version.

Another option could be to port to python, they have working qt bindings.

pierrewillenbrockdfki:
There are at least three robot control systems i am aware of: syskit, rock-cpps state_machine+libinit and the system currently used by DFKI, where the software has not been published yet(a glimpse of which can be found using this search: "Wirkus 2020 Article Online Reconfiguration Of Distributed Robot Control Systems"), which is also built on top of bits from rock-cpp. I believe you are currently developing a fourth system. These systems are capable of coexisting to some extent in the same robot.

It is possible to select the right combination of packages in the buildconf to give you the robot control system of choice. There is no need to have a configuration option for that alone.

Now, one could force the orocos-cpp users to qt5, and the syskit users to qt4, but what about the syskit users that can cope with not being able to use qtruby and vizkit, or the orocos-cpp users that like the ruby rock-display.

Anyway the Qt version configuration option will have to exist in some name; from the point of view of packages, that is the dividing line most packages are interested in, mainly to compile their vizkit3d plugins against the right qt library(source does not differ much). (Thinking about it, one could also look into building both qt4 and qt5 variants and install each in different directories. That would have to be supported by changes to the rock-base/cmake modules, but would remove the need for a Qt version configuration option, at least for libraries/plugins. rock-simulation/mars is not going to like that, though.)

Also, while discussing these topics is interesting and probably useful, i was rather looking for comments like "QTVER is not descriptive enough", or "use this 1 neat trick to have your packages use the branch chosen by configuration, but have that not spill into other unrelated packages/_sets".

doudou:
> I believe you are currently developing a fourth system.

Nope. 100% committed to Syskit here. Just see the trail of commits.

> It is possible to select the right combination of packages in the buildconf to give you the robot control system of choice. There is no need to have a configuration option for that alone.

As you point it out yourself, it does not. If the UI of choice for your control system uses Qt5, then it can't coexist meaningfully with e.g. Syskit. Plus, the "fork" hardly is only about Qt. It also exists on robot description (Syskit uses gazebo and we focus on SDF, you guys are still using URDF + your layer on top I believe).

Why I am having this discussion is: if you go for Qt5, then vizkit is broken, then Syskit's runtime monitoring is broken. Ergo, you can't have the rock.core package set "as it is" with Qt5.

Given that "your" toolchains are neither public nor documented, I would personally go for having the QT version to 4 by default (no questions asked) and you guys have your toolchains package sets/buildconf set it to 5 by default. But I do recognize that I am biased, which is why I suggested we simply split have rock.core with both in parallel and create different package sets with different default settings.

rock.base -> all common packages
rock.syskit
rock.cpp
rock.whatever-the-name-of-your-internal-tool-is

with rock.core being rock.base with Qt4

qtbindings have been abandoned since ... forever so I would not count of them. The path of least resistance would be to migrate the Ruby-Qt things to QML which is more limited, but easy to get to work cross-languages.

pierrewillenbrockdfki:
So how would this actually work:

* we have packages that can use either qt4 or qt5(but those must be consistent to a certain degree, mostly due to vizkit3d) and don't really care about the control system (which are (most of?) the drivers, control, learning, ... defined in their respective package_sets('rock', 'rock.learning', ...)
* the above package_sets ideally would all reference the package_set that defines the qt4 _and_ qt5 osdeps, and that (pre-)defines some configuration entry that tells them which of qt4/qt5 they are supposed to use
* One can then pull another package_set into their buildconf that either offers a choice, or just overrides the configuration entry to whatever it likes best

Does that sound about right, or do we also need to fork the package_sets that currently import 'rock_core', leaving basically 'orocos.toolchain' in common?

There may be a bit of trouble around feature/qt5 branches that are helpful for getting a testable system running quicker, which may not be possible to keep for packages defined alongside the qt[45] osdeps. For those, i think that is fine, they should go away asap. I would like to keep the _option_ for other packages.

pierrewillenbrockdfki:
> So how would this actually work:
> 
>     * we have packages that can use either qt4 or qt5(but those must be consistent to a certain degree, mostly due to vizkit3d) and don't really care about the control system (which are (most of?) the drivers, control, learning, ... defined in their respective package_sets('rock', 'rock.learning', ...)
> 
>     * the above package_sets ideally would all reference the package_set that defines the qt4 _and_ qt5 osdeps, and that (pre-)defines some configuration entry that tells them which of qt4/qt5 they are supposed to use
> 
>     * One can then pull another package_set into their buildconf that either offers a choice, or just overrides the configuration entry to whatever it likes best

Looks like that actually works, even keeping the vizkit3d branch override.

For vizkit3d, i am also planning to change the installation location of vizkit3d plugins that have been compiled against qt5. First of all, that allows parallel installation of both qt4 and qt5 versions, without tooling picking up plugins they cannot work with. Combining that with a (to be done) ability of rock-cmake-macros to work with both qt versions at the same time, one can even install both variants of the same vizkit3d plugin if both qts are available. The QTVER would primarily be used by packages that produce binaries or that are not yet able to build both libraries. Only issue i see here are the vizkit3d debug drawings, they somehow seem to pull qt into orocos transport/marshalling libraries, which obviously cannot be used by processes that use the other qt. Imo, they should not do that, and i hope we don't have to change API to fix that.

planthaber:
> Only issue i see here are the vizkit3d debug drawings, they somehow seem to pull qt into orocos transport/marshalling libraries

I once had a look in another context, the reason is that orogen does not differentiate between dependencies for the task and dependencies for the transport plugin. 

@doudou Perhaps is having a task only version of "using_library", "using_task_library" an option?

pierrewillenbrockdfki:
> Only issue i see here are the vizkit3d debug drawings, they somehow seem to pull qt into orocos transport/marshalling libraries, which obviously cannot be used by processes that use the other qt. Imo, they should not do that, and i hope we don't have to change API to fix that.

There actually was an opportunity for a clean interface already, that separates the qt using parts from the parts that are required for the transports and typekits. It just was not realized, and now it is.

So i explored what is required to produce separate libraries for qt4 and qt5, and the major changes are in base/cmake, see https://github.com/rock-core/base-cmake/pull/79 . Vizkit3d plugin libraries get installed into the plain lib folders, so their name must differ for the plugins to be found. which is also required by cmake to build them in the same project. vizkit3d already scans for -viz.so, so i made the qt5 version scan for -viz-qt5.so. This requires changes to many of the feature/qt5 branches, but i think the results are worth it.

pierrewillenbrockdfki:
This will not be merged in this form, so closing it. There will be smaller incremental PRs for the bits needed, and some rethinking about how to tackle the manifest.xml dependencies if a user decides they don't need one of qt5 or qt4 even though their OS offers them.

PR #218, opened 2023-01-23
Feature/qt5only

pierrewillenbrockdfki:
This adds qt5_designer and osg_qt5 and changes the hard-coded depend of vizkit3d to osg_qt5.

Different to my other PR, this does not add back qt4 and does not change branches of preexisting packages.

@planthaber , this should be uncontroversal enough to be merged onto feature/qt5

### DIFF
--- a/00new_packages.autobuild
+++ b/00new_packages.autobuild
@@ -33,3 +33,6 @@ end
 
 cmake_package "gui/osg_qt4"
 remove_from_default "gui/osg_qt4"
+
+cmake_package "gui/osg_qt5"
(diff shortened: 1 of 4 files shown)


planthaber:
Is there a reason why the osg_qt5 is modeled in the autobuild file and not in the manifets.xml of the package?

pierrewillenbrockdfki:
> Is there a reason why the osg_qt5 is modeled in the autobuild file and not in the manifets.xml of the package?

No, and this should IMO go away as well. But for the moment, it is like it is.

PR #219, opened 2023-04-05
chore: use a "forked" gem for qtbindings instead of using git directly

doudou:
This way we can use caching methods for it instead of having to use unreliable methods (like pre-building it in the CI docker containers)


doudou:
Ping

afonsofonbraga (approved):

PR #227, opened 2024-02-16
feat: implement a "rubocop version manager"

doudou:
This is a script that auto-installs and auto-starts different version of rubocop based on a .rubocop-version file in the packages. This is meant as a way to smoothly migrate away from rubocop 0.83.0 ... which we are stuck on for a long while now.

The script acts only when 'rubocop' gets called.


rubocop-bin/rubocop
```diff
@@ -0,0 +1,82 @@
...
+version_dir = File.join(root, "versions", version)
+gemfile = File.join(version_dir, "Gemfile")
+init_gemfile = File.read(init_gemfile_path) if init_gemfile_path
+
+old_init_gemfile_path = File.join(version_dir, "Gemfile.user")
+old_init_gemfile = File.read(old_init_gemfile_path) if File.exist?(old_init_gemfile_path)
+
+if !File.directory?(version_dir) || (old_init_gemfile != init_gemfile) || (version == "latest")
+    begin
+        STDERR.puts "Auto-installing bundler v#{version}"
```
jhonasiv, line 45:
```suggestion
        STDERR.puts "Auto-installing rubocop v#{version}"
```

jhonasiv (approved):

PR #230, opened 2024-05-15
Remove hard coded osg_qt* dependency for gui/vizkit3d

pierrewillenbrockdfki:
This removes the gui/osg_qt4 dependency of gui/vizkit3d that is added in the 04stable.autobuild file. The manifest.xml of gui/vizkit3d has carried this dependency now for quite some time, so lets remove it here.


pierrewillenbrockdfki:
I am planning to merge this in about a week, if no one objects/merges this quicker.

PR #231, opened 2024-05-23
Add qt5 osdeps

pierrewillenbrockdfki:
Lets start with the simple bits: The osdeps.

I tried my best to find the package names for fedora, opensuse, macos-port, macos-brew, arch and manjarolinux, but i could not confirm that those are the packages i am looking for.

For qt5, we are looking for: headers and libraries for gui, widgets, core, svg, uiplugins and the tools(moc, uic, rcc).

The qt5-designer is used to pull the tool itself, and the libraries/headers, if they don't come with the uiplugins package.

### DIFF
--- a/rock.osdeps
+++ b/rock.osdeps
@@ -136,6 +136,15 @@ qt4:
     macos-port: qt4-mac
     arch,manjarolinux: qt4
 
+qt5:
+    debian,ubuntu: [qtbase5-dev, qtbase5-dev-tools, libqt5svg5-dev, qttools5-dev]
+    gentoo: ["dev-qt/qtcore:5", "dev-qt/qtgui:5", "dev-qt/qtsvg:5"]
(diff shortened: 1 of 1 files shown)


pierrewillenbrockdfki:
@doudou @planthaber @2maz @malter any of you have some input for the "more exotic" operating systems?

I have two more commits i'd like to land in this package-set before we can start transitioning in the qt5 vizkit3d stuff.

One is to add an external source for qtpropertybrowser, and one is to add a source for the qt5 variant of the osg viewer. I will do separate merge requests for each of them, since i imagine some people might have ideas to improve those two.

https://github.com/pierrewillenbrockdfki/rock_core_package_set/tree/feature/qt4-qt5-cleanedup

planthaber:
IMO the osdeps for "exotic" OS should be a "community effort" and provided by those using them.

PR #232, opened 2024-05-24
Replace File.exists with File.exist

malter:


pierrewillenbrockdfki:
ruby 2.5 has both exist? and exists?, ruby 3.0 removed exists?.

PR #236, opened 2024-06-18
Add gui/qtpropertybrowser for gui/vizkit3d

pierrewillenbrockdfki:
This is included in gui/vizkit3d for qt4, but was externalised for qt5.

### DIFF
--- a/04stable.autobuild
+++ b/04stable.autobuild
@@ -217,6 +217,8 @@ in_flavor 'master', 'stable' do
         pkg.env_add_path 'QT_PLUGIN_PATH', File.join(pkg.prefix, "lib", "qt")
     end
 
+    cmake_package 'gui/qtpropertybrowser'
+
     ##### Base implementation for data acquisition
     cmake_package 'drivers/iodrivers_base'
     orogen_package 'drivers/orogen/iodrivers_base'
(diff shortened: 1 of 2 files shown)


pierrewillenbrockdfki:
This would use the repository abhijitkundu/QtPropertyBrowser, but i am not sure if that is a good idea.

Options i see:
1.  Use abhijitkundu/QtPropertyBrowser
2. Upgrade the QtPropertyBrowser in vizkit3d for qt5(possibly using abhijitkundu/QtPropertyBrowser)
3. Extract QtPropertyBrowser from vizkit3d, port to qt5(possibly using abhijitkundu/QtPropertyBrowser)

I think the least messy variant would be one where the same package provides both qt4 and qt5 variants, at least from a dependency point of view. I seem to remember that there are other users of QtPropertyBrowser waiting, so i am leaning towards extracting it from vizkit3d.

pierrewillenbrockdfki:
Chose option 3, the repository is here: https://github.com/rock-core/gui-QtPropertyBrowser. It does provide cmake targets and pkg-config for both qt4 and qt5.

I updated the repositories looking for QtPropertyBrowser to also look for QtPropertyBrowser-qt5(that is only ever the case on the feature/qt5 and feature/qt4-qt5 branches). I tested that the mere presence of gui/QtPropertyBrowser does not disturb the master branch version of vizkit3d.

I'll leave this MR open for a little bit longer, probably about a week.

PR #239, opened 2024-08-14
feat: control the new ROCK_USE_SANITIZERS cmake define using a cxx_sanitizers option

doudou:
Depends on: (weakly)
- [ ] https://github.com/rock-core/rtt/pull/11
- [ ] https://github.com/rock-core/tools-typelib/pull/10
- [ ] https://github.com/rock-core/base-cmake/pull/90

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,26 @@ of `autoproj/manifest`:
 package_sets:
 - github: rock-core/package_set
 ```
+## Configuration options
+
+### `cxx_sanitizers` (EXPERIMENTAL)
+
+`cxx_sanitizers` can be set to a comma separated list of sanitizers which will
+be passed to gcc's -fsanitize= option, when using the Rock.cmake macro package.
+If the list contains 'address', `libasan_path` must also be provided, pointing to
+the absolute path to the libasan library.
+
+The configuration is not fine-grained. All programs will run under libasan as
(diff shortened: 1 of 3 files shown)


pierrewillenbrockdfki:
Technically, this looks good to me. I was initially confused why the `LSAN_OPTIONS` come into play when using `address`, but the `leak` documentation cleared that up. The gcc info page should have contained that tidbit, also the `address` sanitizer wiki page, but here we are.

I would like to see some documentation(probably just a comment if you don't plan to ask for the configuration):
```
# cxx_sanitizers can be set to a comma separated list of sanitizers which will be passed
# to gccs -fsanitize= option, when using the Rock.cmake macro package.
# If the list contains 'address', libasan_path must also be provided.
```

Also, a rationale for setting LSAN_OPTIONS and detect_leaks=0:
```
# On x86_64, the address sanitizer may enable a leak checker
# https://github.com/google/sanitizers/wiki/AddressSanitizerLeakSanitizer
# disable it here and make sure it does not change exit_code
```

pierrewillenbrockdfki (approved):

doudou:
Thanks for the review @pierrewillenbrockdfki. I actually had to completely disable leak checks for the whole workspace, as it creates a lot of spurious command output.

I documented everything in the README. Including the fact that it currently simply does not work.

wvmcastro (approved):

PR #245, opened 2024-11-11
chore: remove flavor-related sanity checks

doudou:
I don't have the time to remove it completely, but that part is safe enough.


planthaber:
Is the idea to remove flavors completely? 

Then I'd also remove the configuration option, and add deprecation warnings to the functions (with instructions to set the ROCK_SELECTED_FLAVOR manually, if required)

https://github.com/rock-core/package_set/blob/master/init.rb#L61-L142

overrides.rb
```diff
@@ -1,41 +1,10 @@
...
-            # Finally, verify that when pkg A from flavor X depends on pkg B,
-            # then B needs to be available in flavor X as well
-            Rock.flavors.verify_cross_flavor_dependencies(pkg)
-        end
-
-        # Do the git hook setup in a separate setup block since we must do it
-        # post-import
-        pkg.post_import do
-            if pkg.importer.branch == "next" || pkg.importer.branch == "stable"
-                Rock.install_git_hook pkg, 'git_do_not_commit_hook', 'pre-commit'
```
pierrewillenbrockdfki, line 32: I guess there is a file "git_do_not_commit_hook" that could also be removed?

planthaber: https://github.com/rock-core/package_set/blob/master/rock/git_do_not_commit_hook

doudou:
> Is the idea to remove flavors completely?

Rather obviously, since 'stable' is pre-2016 I think ... or even earlier.
 
> Then I'd also remove the configuration option, and add deprecation warnings to the functions (with instructions to set the ROCK_SELECTED_FLAVOR manually, if required)
> 
> https://github.com/rock-core/package_set/blob/master/init.rb#L61-L142

As I said. I don't have time to do the whole removal. If you volunteer, be my guest.

overrides.rb
doudou: Same answer.

Some stuff removed is better than nothing removed. Nobody cared for something like 10 *years*. If you guys do have time, be my guest.

pierrewillenbrockdfki (approved):
Looks good to me.

planthaber:
Do we actually need deprecation?

I think we could also just remove the functionality, as it wasn't used for a long time.

planthaber (approved):

jhonasiv (approved):

doudou:
> Do we actually need deprecation?
> 
> I think we could also just remove the functionality, as it wasn't used for a long time.

Agreed. I did not add deprecations...

PR #247, opened 2024-11-14
Osdeps qwt5 osg_qt4

pierrewillenbrockdfki:
Mark qwt5 and osg_qt4 non-existent on more platforms

### DIFF
--- a/rock.osdeps
+++ b/rock.osdeps
@@ -124,6 +124,7 @@ gui/osg_qt4:
     ubuntu:
         '16.04,18.04,18.10,19.04,19.10':
             osdep: osg
+        default: nonexistent
 
 qt4:
     # QMake is needed for the CMake macro for Qt4
--- a/rock.osdeps
+++ b/rock.osdeps
@@ -81,7 +81,7 @@ rice:
 qwt5:
     debian,ubuntu:
         default: libqwt5-qt4-dev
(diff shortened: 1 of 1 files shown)


pierrewillenbrockdfki:
These are the bits that are left when i remove the uneeded stuff from https://github.com/pierrewillenbrockdfki/rock_core_package_set

rock.osdeps
```diff
@@ -81,7 +81,7 @@ rice:
 qwt5:
     debian,ubuntu:
         default: libqwt5-qt4-dev
-        '20.04,22.04': nonexistent
+        '20.04,22.04,24.04': nonexistent
```
planthaber, line 84: This should be switched around, default should be the "latest" osdep (in hope this does not change in future).

please use  default: nonexistent and state os for libqwt5-qt4-dev explicitely

pierrewillenbrockdfki:
> Regarding QWT, there is actually another PR #202 which suggests to install `libqwt-qt5-dev` instead of `libqwt5-qt4-dev`. I don't have any packages with a non-optional dependency on QWT, so either solution is fine for me.

qwt5 is not about qt5 support, it is about qwt5 support; qwt5 generally is compiled for qt4, while qwt6 generally is compiled for qt5.

chhtz (approved):

PR #250, opened 2025-03-21
chore: remove dependency on the orocos.toolchain package set

doudou:


jhonasiv (approved):

pierrewillenbrockdfki (commented):
There are packages in rock.core that require ruby-dev from orocos.toolchain(`binding_of_caller`, `rice`, `thin`, `qtruby`, `concurrent-ruby`). You probably want to move the osdep definition of ruby-dev in here as well.

PR #251, opened 2025-09-04
Define `RUBY_EXECUTABLE` and `Ruby_EXECUTABLE` for packages which need it

chhtz:
That line breaks cmake configuration for some packages on Ubuntu 22.04, since CMake changed that variable to `Ruby_EXECUTABLE`.

Related: https://github.com/rock-core/base-cmake/pull/94

@doudou Do you know if there is a reason `RUBY_EXECUTABLE` was passed manually to some packages?

### DIFF
--- a/04stable.autobuild
+++ b/04stable.autobuild
@@ -5,6 +5,12 @@ manifest = if Autoproj.respond_to?(:workspace)
 
 add_packages_to_flavors 'stable' => ['orogen', 'rtt', 'utilmm', 'utilrb', 'typelib', 'rtt_typelib', 'tools/metaruby', 'ocl', 'log4cpp']
 
+def set_cmake_ruby_executable(pkg)
+    pkg.define "Ruby_EXECUTABLE", Autoproj.config.ruby_executable
+    # This is for compatibility with Ubuntu 20.04/CMake 3.16
+    pkg.define "RUBY_EXECUTABLE", Autoproj.config.ruby_executable
+end
+
 in_flavor 'master', 'stable' do
     ruby_package 'gui/vizkit'
 
@@ -34,7 +40,7 @@ in_flavor 'master', 'stable' do
     remove_from_default "bundles/rock"
     cmake_package 'tools/pocolog_cpp'
     cmake_package 'tools/service_discovery' do |pkg|
-        pkg.define "RUBY_EXECUTABLE", Autoproj.config.ruby_executable
+        set_cmake_ruby_executable pkg
     end
 
     ruby_package 'tools/autoproj'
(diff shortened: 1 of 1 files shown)


chhtz:
According to [this link](https://cmake.org/cmake/help/v3.18/module/FindRuby.html), CMake's FindRuby actually will *provide* both `Ruby_THING` and `RUBY_THING` (the latter is discouraged to be used and deprecated since CMake 4.0)/ The problem is that some versions of FindRuby struggle if `RUBY_EXECUTABLE` is provided via `-DRUBY_EXECUTABLE=ruby_path`, giving a weird error like:
```
    -- Could NOT find Ruby (missing: Ruby_EXECUTABLE) (found suitable version "3.0.2", minimum required is "1.8.0")
```

 (in Ubuntu 24.04, CMake 3.28.3, this seems not to be a problem anymore).

If there is a reason to pass `Autoproj.config.ruby_executable`, we need either check if cmake_version < 3.18 and provide `RUBY_EXECUTABLE` or `Ruby_EXECUTABLE` accordingly, or always pass both.
If there is no reason to manually pass `Ruby_EXECUTABLE`, I prefer to remove it.

doudou (commented):
These guarantee that all FindRuby find the same ruby executable. It is needed.

chhtz:
> These guarantee that all FindRuby find the same ruby executable. It is needed.

Ok, fair enough. (Even though, I guess always using the default should be the same everywhere as well, I guess you have use cases where you want to use the non-default ruby version).

I will work on another patch.

chhtz:
I will update https://github.com/orocos-toolchain/autoproj/pull/40 accordingly, once this gets merged

doudou (approved):

chhtz:
> Wouldn't "set_cmake_ruby_executable" be a better name for the function. By just reading "use_ruby_executable pkg" in the package definition, i can't really figure out what it is supposed to do.

I'm completely open regarding the name of the function ...

> > Also, if there is an elegant way to determine the CMake version which is in use, one could define only the correct variable.
> 
> https://stackoverflow.com/questions/43697154/detect-current-cmake-version-using-cmake:
> 
> ```ruby
> if(CMAKE_VERSION VERSION_LESS "3.8.0") 
>     message("Please consider to switch to CMake 3.8.0")
> endif()
> ```
> [...]

But this only works inside cmake, not from ruby, right? I could execute `cmake --version` or something and evaluate the result, but IMO that is over-complicating the issue (unless this is done by autoproj already and I can just take the result -- but a quick search did not give any results for me).

chhtz:
@planthaber I changed the macro name now. Do you want to merge?

PR #252, opened 2025-09-24
Create Python venv instead of manual site-packages

chhtz:
Work in progress


chhtz:
This is in a mostly working state, some possible improvements:

* [ ] Code should be moved to autoproj directly, all functions in this package set should call the corresponding autoproj function and get marked deprecated
* [ ] Code generally could be tested more, especially what happens when deactivating/re-activating
  * [ ] Make sure no illogical options are possible (activate venv without activating python)
  * [ ] Also, what happens if venv is (partially) deleted manually, does it recover by an `autoproj osdep`?
* [ ] Make sure no downloads/installation happens, except in the osdeps-phase
* [ ] Location of venv could be made configurable (in case one wants to share it between workspaces -- only works if installed versions are identical)

PR #254, opened 2026-04-20
fix: add qt4-related definitions for Ubuntu 24.04

doudou:
I could not get to build webkit, so removed it from the list (and will remove its usage in the ruby toolchain ...)


jhonasiv (approved):

pierrewillenbrockdfki (approved):
Assuming that you compile tested this, i approve.

I cannot seem to find the source you used to build the qt4 libraries, do you mind sharing that? I must have known at some point but cannot remember where to look.

doudou:
Since it is on a PPA, the deb source packages are available (it is mandatory to let the PPA actually build from source)

